PR #1, opened 2019-10-03
Refactor the spec to actually send the origin to the server

vasilvv:


draft-vvv-webtransport-quic.md
```diff
@@ -96,7 +96,9 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
 when, and only when, they appear in all capitals, as shown here.
 
-This document follows terminology defined in Section 1.2 of [OVERVIEW].
+This document follows terminology defined in Section 1.2 of [OVERVIEW].  The
+diagrams describe encoding following the conventions described in Section 1.3 of
+[QUIC-TRANSPORT].
```
DavidSchinazi: To get this to turn into a link, I think you want `{{QUIC-TRANSPORT}}` with a definition similar to what's in draft-ietf-quic-tls.

draft-vvv-webtransport-quic.md
```diff
@@ -110,10 +112,10 @@ beneficial.
 
 When a client requests a QuicTransport to be created, the user agent establishes
 a QUIC connection to the specified address.  It verifies that the the server is
-a QuicTransport endpoint using ALPN, and that the client is allowed to connect
-to the specified endpoint using `web_accepted_origins` transport parameter.
-Once the verification succeeds and the QUIC connection is ready, the client can
-send and receive streams and datagrams.
+a QuicTransport endpoint using ALPN, and sends a *client indication* containing
```
DavidSchinazi: Is endpoint a defined term in IETF land?

draft-vvv-webtransport-quic.md
```diff
@@ -129,23 +131,70 @@ the client receives a TLS Finished message from the server.
 
 In order to identify itself as a WebTransport application, QuicTransport relies
 on TLS Application-Layer Protocol Negotiation {{!RFC7301}}.  The user agent MUST
-request the ALPN value of "wq" and it MUST NOT establish the session unless that
-value is accepted.
+request the ALPN value of "wq-draft01" and it MUST NOT establish the session
```
DavidSchinazi: I'd personally prefer `wq-01` to match the QUIC docs

DavidSchinazi: I'm confused by this MUST NOT establish - if the server rejects the value it fails the handshake and closes the connection right?

draft-vvv-webtransport-quic.md
```diff
@@ -129,23 +131,70 @@ the client receives a TLS Finished message from the server.
...
+unless that value is accepted.
+
+## Client Indication
+
+In order to verify that the client's origin is allowed to connect to the server
+in question, the user agent has to communicate the origin to the server.  This
+is accomplished by sending a special handshake message, called *client
```
DavidSchinazi: I wouldn't call this a handshake message because it could be interpreted as meaning QUIC/TLS handshake

draft-vvv-webtransport-quic.md
```diff
@@ -129,23 +131,70 @@ the client receives a TLS Finished message from the server.
...
+stream.
+
+The client indication is a sequence of key-value pairs that are formatted in the
+following way:
+
+~~~~~~~~~~ drawing
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                           Key (2)                           ...
```
DavidSchinazi: QUIC notation is in bits, not bytes. Also, can't we make these varints?

draft-vvv-webtransport-quic.md
```diff
@@ -129,23 +131,70 @@ the client receives a TLS Finished message from the server.
...
+
+## Client Indication
+
+In order to verify that the client's origin is allowed to connect to the server
+in question, the user agent has to communicate the origin to the server.  This
+is accomplished by sending a special handshake message, called *client
+indication*, on stream 2, which is the first client-initiated unidirectional
+stream.
+
+The client indication is a sequence of key-value pairs that are formatted in the
```
DavidSchinazi: We probably need a way to convey the end of the sequence. For example
- number of pairs up front
- total length up front
- FIN sent after the last one

draft-vvv-webtransport-quic.md
```diff
@@ -129,23 +131,70 @@ the client receives a TLS Finished message from the server.
...
+  Length:
+  : Indicates the length of the value (the length of the key and the length
+    itself are not included).
 
-## Verifying the Origin
+  Value:
+  : The value of the field, the semantics of which are determined by the key.
+
+The total length of the client indication MUST NOT exceed 65,535 bytes.  The
+keys MAY be duplicate, depending on the definition of the field.  The server
```
DavidSchinazi: It's probably simplest to ban duplicate keys like QUIC transport parameters, as this flexibility adds risk

draft-vvv-webtransport-quic.md
```diff
@@ -129,23 +131,70 @@ the client receives a TLS Finished message from the server.
...
+
+  Key:
+  : 0x0000
 
-In order to verify that the client is authorized to access a specific
-WebTransport server, QuicTransport has a mechanism to verify the origin
-{{!RFC6454}} associated with the client.  The server MUST send a
-`web_accepted_origins` transport parameter which SHALL be one of the following:
+  Description:
+  : The origin {{!RFC6454}} of the client initiating the connection.
```
DavidSchinazi: How is the origin encoded? (see RFC6454 section 6)

draft-vvv-webtransport-quic.md
vasilvv: It's already a link, though.

draft-vvv-webtransport-quic.md
vasilvv: Well, quic-transport-23 defines it and uses it 460 times.

draft-vvv-webtransport-quic.md
vasilvv: I actually changed it to `wq-vvv-01`, since this is not a WG draft.

vasilvv: I changed it to "MUST close".  FWIW, this is just a restatement of quic-transport requirement.

draft-vvv-webtransport-quic.md
vasilvv: Removed word "handshake".

draft-vvv-webtransport-quic.md
vasilvv: Fixed the bytes one.  I've considered making those varints, but my concern was that not all QUIC libraries expose the varint parser to the applications, thus making this potentially harder to implement.

draft-vvv-webtransport-quic.md
vasilvv: It's meant to be FIN.  Spelled this out more clearly.

draft-vvv-webtransport-quic.md
vasilvv: Done.

draft-vvv-webtransport-quic.md
vasilvv: I already say that below.

draft-vvv-webtransport-quic.md
DavidSchinazi: Huh, weird

draft-vvv-webtransport-quic.md
DavidSchinazi: SGTM

draft-vvv-webtransport-quic.md
DavidSchinazi: Varints take about 10 lines of code to implement, I think they're worth the flexibility

draft-vvv-webtransport-quic.md
DavidSchinazi: Where? Section 6 of RFC6454 proposes multiple different encodings

vasilvv: Ah, good point.  Clarified that.

draft-vvv-webtransport-quic.md
```diff
@@ -129,23 +137,73 @@ the client receives a TLS Finished message from the server.
...
+
+~~~~~~~~~~ drawing
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                           Key (16)                          ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                          Length (16)                        ...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                           Value ( )                         ...
```
flano-yuki: I think variable length fields are indicated by an asterisk
(as https://quicwg.org/base-drafts/draft-ietf-quic-transport.html#notational-conventions

flano-yuki:
What is the uri-scheme used in an origin?
( It may be discussed as a separate issue.

draft-vvv-webtransport-quic.md
vasilvv: Indeed.  Thanks for pointing this out.

vasilvv:
> What is the uri-scheme used in an origin?
> ( It may be discussed as a separate issue.

Well, the origin sent here is the origin of the webpage that created QuicTransport; so, for instance, if `https://website.example/webapp` connects using QuicTransport to `quic-server.example:443`, the origin  send would be `https://website.example`.

I don't think a QuicTransport connection can be an origin by itself.  We are currently considering giving it its own URL scheme, but a URL can only act as an origin if a user can [navigate](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigate) to it, or if the URL has storage attached to it (which QuicTransport doesn't, since it doesn't support HTTP cookies).

flano-yuki:
I misunderstood the URI schema as that of QuicTransport itself 🙇 
Your kind explanation made me understand.

thanks

PR #7, opened 2019-11-01
dschinazi editorial pass on overview

DavidSchinazi:


draft-vvv-webtransport-overview.md
```diff
@@ -349,16 +356,14 @@ the network addresses that would normally be inaccessible.  Because of that, the
 user agent that runs untrusted clients MUST NOT provide any detailed error
 information until the server has confirmed that it is a WebTransport endpoint.
 For example, the client must not be able to distinguish between a network
-address that is unreachable and that is reachable but is not a WebTransport
+address that is unreachable and one that is reachable but is not a WebTransport
 server.
 
-WebTransport does not support any traditional means of browser-based
+WebTransport does not support any traditional means of HTTP-based
```
vasilvv: I think "HTTP-based authentication" does not actually include all of the things that we want to not support (e.g. prompt-based client auth).

draft-vvv-webtransport-overview.md
```diff
@@ -296,46 +302,47 @@ the size of the congestion window or the current pacing rate.
...
 # Buffering and Prioritization
 
 TODO: expand this outline into a full summary.
 
 * Datagrams are intended for low-latency communications, so the buffers for them
   should be small, and prioritized over stream data.
-* In general, the transport should not use any aggregation algorithms, e.g.
-  Nagle's algorithm {{?RFC0896}}.
+* In general, the transport should not apply aggregation algorithms (e.g.,
+  Nagle's algorithm {{?RFC0896}}) to datagrams.
```
vasilvv: I was actually mostly worried about stream data.

draft-vvv-webtransport-overview.md
```diff
@@ -139,8 +147,8 @@ Message:
 Transport feature:
 
 : A transport feature refers to the ability of a specific transport to
-  provide a specific way of communicating data, such as supporting datagrams or
-  streams.
+  provide a specific mechanism to communicate data.  An example is the ability
+  to send datagrams that are not retransmitted in case of packet loss.
```
vasilvv: That's a transport property.  The feature I was trying to define here are "supports streams" and "support datagrams".

draft-vvv-webtransport-overview.md
DavidSchinazi: What do you mean by "prompt-based client auth" ? What non-HTTP-based authentications are there?

draft-vvv-webtransport-overview.md
DavidSchinazi: Nagle has value, I don't think we should argue against Nagle without a more detailed explanation

draft-vvv-webtransport-overview.md
DavidSchinazi: I don't understand the difference

draft-vvv-webtransport-overview.md
vasilvv: TLS client auth (the one where user gets a prompt).

draft-vvv-webtransport-overview.md
vasilvv: Nagle does have value, and I'm pretty sure it's negative for ~all latency-sensitive applications we're targeting with this, at least without explicit controls.

draft-vvv-webtransport-overview.md
vasilvv: It's a difference between "what you can do with transport" and "how the transport will do it".  I agree it's not clear, so I just deleted that definition from the list.

draft-vvv-webtransport-overview.md
DavidSchinazi: Not all streams are latency-sensitive in a latency-sensitive application though

draft-vvv-webtransport-overview.md
DavidSchinazi: WebTransport runs over TLS or QUIC so it has no reason not to support TLS client auth?

Issue #15, opened 2019-12-04
Define Error Code of WebTransport over QUIC


flano-yuki:
In WebTransport over QUIC, when an invalid  Client Indication is received, the server	 treats the connection as an error([section 6](https://vasilvv.github.io/webtransport/draft-vvv-webtransport-quic.html#rfc.section.6)).

This is an error that occurred in a layer above the QUIC, so I think the connection should be closed with CONNECTION_CLOSE (0x1d) indicating an error in the application using the QUIC.

I think it is valuable to define an error code for WebTransport over QUIC.
( [draft-ietf-quic-transport-24 section 20.1](https://tools.ietf.org/html/draft-ietf-quic-transport-24#section-20.1) saids "the management of application error codes is left to application protocols." )

and, by specifying the error code that the endpoint uses when it close the connection, the specification becomes clear.

for example
```
- WQ_NO_ERROR 
- WQ_INVALID_CLIENT_INDICATION
- WQ_NOT_AUTHORITATIVE_ORGIN
- WQ_OTHER_ERROR
...
```

flano-yuki:
I realized that Error Code is related to the [WICG/web-transport](https://github.com/WICG/web-transport).

I createted https://github.com/WICG/web-transport/issues/80 .
Please point it out if there are any mistakes in my opinion.

kixelated:
I agree, CONNECTION_CLOSE should provide an error code indicating why the connection was refused.

LPardue:
I seem to remember seeing non-zero error codes from other endpoints when interoperating. If so, I have no clue what those codes were supposed to mean. I think it would be useful to have some defined in the next draft version.

vasilvv:
I think the non-zero error codes probably were random HTTP/3 error codes, because those are the only codes that Google's QUIC library supports (that's a known bug).

I see three options for reporting handshake error:
1) An application CONNECTION_CLOSE,
2) A RESET_STREAM on the handshake stream,
3) An error status code on the handshake stream.

If we go with 1, we would have to reserve some of the application error code space.  If we go with 2 or 3, we would have to describe how the server has to keep the connection open for some time in order to allow for retransmission of the RESET_STREAM/response error header.

DavidSchinazi:
Closing this issue because the QuicTransport draft has been removed from this repository.

PR #18, opened 2020-09-04
Refactor client indication into a full header exchange

vasilvv:
We currently use 16-bit integers for header names.  There probably won't
ever be more than 65,000 headers, but this does not let users add their
own custom headers easily.  Thus, this commit changes uint16 to string.

We currently provide a way for a server to explicitly reject client's
origin, but no way to explicitly accept it.  This commit solves this by
turning the client indication stream into a bidirectional handshake
stream.  Server confirms the handshake by sending a FIN on its end of
the handshake stream.

### DIFF
--- a/draft-vvv-webtransport-quic.md
+++ b/draft-vvv-webtransport-quic.md
@@ -139,10 +139,10 @@ be used in situations where these features are desired.
 
 When a client requests a QuicTransport session to be created, the user agent
 establishes a QUIC connection to the specified address.  It verifies that the
-the server is a QuicTransport endpoint using ALPN, and additionally sends a
-client indication containing the requested path and the origin of the initiating
-website to the server.  At that point, the connection is ready from the client's
-perspective.  The server MUST wait until the client indication is received
+the server is a QuicTransport endpoint using ALPN, and additionally sends a set
+of headers containing the requested path and the origin of the initiating
+website to the server.  At that point, the connection is ready from the
+client's perspective.  The server MUST wait until it processes client headers
 before processing any application data.
 
 WebTransport streams are provided by creating an individual unidirectional or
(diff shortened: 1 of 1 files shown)


draft-vvv-webtransport-quic.md
```diff
@@ -154,62 +154,70 @@ datagram extension [QUIC-DATAGRAM].
...
+header block size.
 
-In order to ensure that the user agent can send the client indication
-immediately, the server MUST set `initial_max_streams_uni` transport parameter
-to at least `1`.  The user agent MUST close the connection if the server sets
-`initial_max_streams_uni` to `0`.
+In order to ensure that the user agent can send the headers immediately, the
+server MUST set `initial_max_streams_bidi` transport parameter to at least `1`.
+The user agent MUST close the connection if the server sets
+`initial_max_streams_bidi` to `0`.
```
LPardue, line 220: Given the bulleted description of the handshake and this section, it is a little unclear what you expect the server to do. For instance, consider a client cannot send the complete block because the server's initial bidi flow control window is too small. Is the expectation that a Server Application will read data from the QUIC stream (giving flow control credit back to client) and buffer the block in the application layer until the client completes? 

It seems unfortunate if a client can force a server to buffer up to 1,048,576 bytes. Yes the server can hangup, but it would have to wait until the application-imposed limit was reached. It might be healthier for the block to begin with a total length field. That allows a server to early abort.

draft-vvv-webtransport-quic.md
```diff
@@ -237,21 +241,20 @@ The user agent MUST send the "Origin" field.  The "Origin" field MUST be set to
 the origin of the client initiating the connection, serialized as described in
 the "serializing a request origin" section of [FETCH].
 
+The server indicates that it accepts the origin by sending its header block and
+a FIN on the stream 0.  If the server does not accept the origin, it MUST close
+the connection.
```
LPardue, line 246: With the current format and defined fields, this means just sending an empty STREAM frame with FIN?

kixelated, line 246: How does the client know that a connection close was due to an unaccepted origin? If it receives a CONNECTION_CLOSE prior to receiving a FIN on stream 0?

The problem is that there are many sources of CONNECTION_CLOSE errors. The server could close the connection due to an invalid path or it just decided to immediately terminate the connection. In either case, the client will receive a QUIC application close and will incorrectly assume that it was due to an invalid origin.

To correctly identify invalid origins, there should be an explicit and retransmitted signal. Either an "invalid origin" response message (on stream 0)  or CONNECTION_CLOSE error code.

LPardue: You could also abort the stream with a RESET_STREAM in this case. That way you wouldn't even need an error code (although it would help) in order to detect the case where the origin exchange is pass/fail.

kixelated: Yeah, `RESET_STREAM` would be nice with or without error codes. It's better than CONNECTION_CLOSE because it can be reliable and does not infringe upon the application error code space.

draft-vvv-webtransport-quic.md
```diff
@@ -154,62 +154,70 @@ datagram extension [QUIC-DATAGRAM].
...
-Key:
-
-: Indicates the field that is being expressed.
-
-Length:
+Name:
 
-: Indicates the length of the value (the length of the key and the length
-  itself are not included).
+: Indicates the name of the field that is being expressed.  The name of the
```
DavidSchinazi, line 200: Why is this more restrictive than HTTP headers? Couldn't we use the same rules as HTTP/2 headers for simplicity?

draft-vvv-webtransport-quic.md
```diff
@@ -154,62 +154,70 @@ datagram extension [QUIC-DATAGRAM].
...
-A FIN on the stream 2 SHALL indicate that the message is complete.  The client
-MUST send the entirety of the client indication and a FIN immediately after
-opening the connection.  The server MUST NOT process any application data before
-receiving the entirety of the client indication.  The total length of the client
-indication MUST NOT exceed 65,535 bytes.
+A FIN on the stream 0 SHALL indicate that the message is complete.  The client
+MUST send the entirety of the header block and a FIN immediately after opening
+the connection.  The server MUST NOT process any application data before
+receiving the entirety of the headers.  The total length of a header block MUST
+NOT exceed 1,048,576 bytes; implementations MAY impose additional limit on the
```
DavidSchinazi, line 214: Additional implementation-specific limits can lead to interop issues. Could we instead say that servers MUST accept headers up to that length?

draft-vvv-webtransport-quic.md
```diff
@@ -403,30 +405,26 @@ a given client.
...
+  : 0x77 0x71 0x2d 0x76 0x76 0x76 0x2d 0x30 0x33 ("wq-vvv-03")
 
   Specification:
   : This document
 
-## Client Indication Fields Registry
+## Header Fields Registry
 
-IANA SHALL add a registry for "QuicTransport Client Indication Fields" registry.
+IANA SHALL add a registry for "QuicTransport Header Fields" registry.
```
DavidSchinazi, line 421: Could we instead use the HTTP headers registry?

DavidSchinazi (commented):
I think that we should stay as close to HTTP headers as possible here, to simplify the relationship between QuicTransport and HttpTransport. We could then say that JavaScript MUST NOT be able to set headers that start with a colon to enforce a separation in the namespace.

draft-vvv-webtransport-quic.md
```diff
@@ -218,17 +226,13 @@ than once.
...
 
 In order to allow the server to enforce its origin policy, the user agent has to
-communicate the origin in the client indication.  This can be accomplished using
-the "Origin" field:
+communicate the origin in the handshake.  This can be accomplished using the
+"Origin" field:
 
 Name:
 
 : Origin
```
yutakahirano, line 234: Shouldn't this be lower-cased?

draft-vvv-webtransport-quic.md
```diff
@@ -237,21 +241,20 @@ The user agent MUST send the "Origin" field.  The "Origin" field MUST be set to
...
 
 In order to allow multiplexing multiple application on the same host-port tuple,
 QuicTransport allows specifying extra routing information in the path component
-of the URI.  That component is communicated using the "Path" field in the client
-indication:
+of the URI.  That component is communicated using the "Path" header:
 
 Name:
 
 : Path
```
yutakahirano, line 256: Shouldn't this be lower-cased?

draft-vvv-webtransport-quic.md
vasilvv: I don't believe anything in the current text of the PR precludes the server from processing headers as they come in, and, say, rejecting based on incorrect origin as soon as it parses the origin.  The requirement is to not _accept_ the headers until they're complete.

We could add a size in front, but this would have a downside of forcing the client to buffer all of the headers to compute the size in advance.

draft-vvv-webtransport-quic.md
vasilvv: Correct.

draft-vvv-webtransport-quic.md
vasilvv: This is actually based on [the most restrictive interpretation of HTTP header name restrictions](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#rfc.section.5.3.p.6).

draft-vvv-webtransport-quic.md
vasilvv: We could, but in practice implementations will inevitable impose their own limit, and we can't preemptively enforce this, so I don't think it makes sense.

draft-vvv-webtransport-quic.md
LPardue: The ambiguity here is " MUST NOT process any application data before receiving the entirety of the headers". Its ambiguous because this is an application mapping and its all application data. I think what you mean is that the server MUST not process any received data on streams other than 0, or process any datagrams.

Based on the defined fields so far, I don't see needing a length upfront as a real problem for clients. HTTP/2 and HTTP/3 require a length in HEADERS. What use case really needs such client behaviour?

draft-vvv-webtransport-quic.md
vasilvv: I'm a bit weary of that because the "Internet Message Header" registry is currently already overloaded, and from what I understand it's currently being reorganized, so I would wait here.

draft-vvv-webtransport-quic.md
vasilvv: I'm following the HTTP convention here where headers look like "Content-Type", even though on the wire, HTTP/2 and HTTP/3 encode those as all-lowercase.

draft-vvv-webtransport-quic.md
vasilvv: I'm following the HTTP convention here where headers look like "Content-Type", even though on the wire, HTTP/2 and HTTP/3 encode those as all-lowercase.

draft-vvv-webtransport-quic.md
LPardue: This seems just like the setting for MAX_HEADER_LIST size. Unfortunately that setting has legacy baggage from decades of HTTP deployment. It seems with a new application protocol, the server could defe a limit thats enforceable. That helps cut down the memory commitment and is interoperable. Exchanging such a setting might require something ALPS-like though.

draft-vvv-webtransport-quic.md
```diff
@@ -154,62 +154,70 @@ datagram extension [QUIC-DATAGRAM].
...
-|           Key (16)            |          Length (16)          |
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-|                           Value (*)                         ...
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+QuicTransport Header {
+  Header Name Length (16),
+  Header Name (..),
+  Header Value Length (16),
+  Header Value (..),
+}
```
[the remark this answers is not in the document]
vasilvv: Oooh, that's an interesting syntax. Done.

draft-vvv-webtransport-quic.md
DavidSchinazi: I do think that we should resolve those potential interop issues - and we can enforce this by using padding right?

draft-vvv-webtransport-quic.md
DavidSchinazi: I'm not sure if I understand why the reorganization of [the registry](https://www.iana.org/assignments/message-headers/message-headers.xhtml) would cause us to not use it

draft-vvv-webtransport-quic.md
DavidSchinazi: I think sticking to HTTP semantics would be best. Adding a reference to that link SGTM as a way to suggest folks use this subset.

draft-vvv-webtransport-quic.md
vasilvv: Re ambiguity, thanks for pointing this out! This should be fixed.

Re header size, I agree we should do something about this, but not sure I like explicit size, given that it's redundant otherwise.  HTTP/2 and HTTP/3 do always have size, but that's because neither sends headers on a dedicated stream, so they need header block size for delineation.  I've filed #20 to track this.

draft-vvv-webtransport-quic.md
vasilvv: I don't think we want to send one megabyte worth of padding.  I've filed #20 to track this.

draft-vvv-webtransport-quic.md
vasilvv: Those are good points.  I wrote my thoughts on them in #15

PR #21, opened 2020-11-09
Update WebTransport over HTTP/3 to be more in sync with HTTP/2

vasilvv:
This replaces a distinct Session ID mechanism with using stream ID of
the CONNECT stream as a Session ID.

The transport parameter used to opt into WebTransport is replaced with
a SETTINGS value.

### DIFF
--- a/draft-vvv-webtransport-http3.md
+++ b/draft-vvv-webtransport-http3.md
@@ -172,8 +172,8 @@ MUST send a SETTINGS_ENABLE_WEBTRANSPORT value set to "1" in their SETTINGS
 frame.  Endpoints MUST NOT use any WebTransport-related functionality unless
 the parameter has been negotiated.
 
-If SETTINGS_ENABLE_WEBTRANSPORT is negotiated, support for the QUIC DATAGRAM
-extension MUST be negotiated as described in
+If SETTINGS_ENABLE_WEBTRANSPORT is negotiated, support for the QUIC DATAGRAMs
+within HTTP/3 MUST be negotiated as described in
 {{!HTTP3-DATAGRAM=I-D.schinazi-quic-h3-datagram}}; negotiating WebTransport
 support without negotiating QUIC DATAGRAM extension SHALL result in a
 H3_SETTINGS_ERROR error.
--- a/draft-vvv-webtransport-http3.md
+++ b/draft-vvv-webtransport-http3.md
@@ -137,17 +137,17 @@ can be accessed via an HTTP/3 server.
 WebTransport servers in general are identified by a pair of authority value and
 path value (defined in {{!RFC3986}} Sections 3.2 and 3.3 correspondingly).
 
-When an HTTP/3 connection is established, the client and server have to
-negotiate a specific set of QUIC transport parameters that indicate support for
-various WebTransport features.  Most notably, the `http3_transport_support`
-parameter signals WebTransport support to the peer.
+When an HTTP/3 connection is established, both the client and server have to
+negotiate a SETTINGS_ENABLE_WEBTRANSPORT setting in order to indicate that they
+both support WebTransport over HTTP/3.
 
 WebTransport sessions are initiated inside a given HTTP/3 connection by the
 client, who sends an extended CONNECT request {{!RFC8441}}.  If the server
-accepts the request, an WebTransport session is established.  As a part of
-this process, the client proposes, and the server confirms, a session ID.
-A session ID (SID) is unique within a given HTTP/3 connection, and is used to
-associate all of the streams and datagrams with the specific session.
+accepts the request, an WebTransport session is established.  The resulting
+stream will be further referred to as a *CONNECT stream*, and its stream ID is
+used to uniquely identify a given WebTransport session within the connection.
+The ID of the CONNECT stream that established a given WebTransport session will
+be further referred to as a *Session ID*.
 
 After the session is established, the peers can exchange data using the
 following mechanisms:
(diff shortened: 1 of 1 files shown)


draft-vvv-webtransport-http3.md
```diff
@@ -163,70 +163,55 @@ following mechanisms:
...
-parameters it intends to send based on whether the client has indicated support
-for WebTransport.
+In order to indicate support for WebTransport, both the client and the server
+MUST send a SETTINGS_ENABLE_WEBTRANSPORT value set to "1" in their SETTINGS
+frame.  Endpoints MUST NOT use any WebTransport-related functionality unless
+the parameter has been negotiated.
+
+If SETTINGS_ENABLE_WEBTRANSPORT is negotiated, support for the QUIC DATAGRAM
+extension MUST be negotiated as described in
+{{!HTTP3-DATAGRAM=I-D.schinazi-quic-h3-datagram}}; negotiating WebTransport
```
DavidSchinazi, line 177: The text mentions QUIC DATAGRAMs but this reference is h3 DATAGRAMS - I think the distinction is important

draft-vvv-webtransport-http3.md
```diff
@@ -325,12 +310,11 @@ variable-length integer, followed by the user-specified stream data
...
-Datagrams can be sent using the DATAGRAM frame as defined in
-[QUIC-DATAGRAM] and {{!H3-DATAGRAM=I-D.schinazi-quic-h3-datagram}}.  For all
-HTTP/3 connections in which the `http3_transport_support` option is negotiated,
-the Flow Identifier is set to the session ID.  In other words, the format of
-datagrams SHALL be the session ID, followed by the user-specified payload
-({{fig-datagram}}).
+Datagrams can be sent using the DATAGRAM frame as defined in [QUIC-DATAGRAM]
+and [HTTP3-DATAGRAM].  For all HTTP/3 connections in which the
+SETTINGS_ENABLE_WEBTRANSPORT option is negotiated, the Flow Identifier is set
+to the session ID.  In other words, the format of datagrams SHALL be the
```
DavidSchinazi, line 316: This goes against the design of draft-schinazi-quic-h3-datagram. Should we use the same mechanism as CONNECT-UDP to allow multiplexing CONNECT-UDP and WebTransport over the same h3 connection?

draft-vvv-webtransport-http3.md
```diff
@@ -413,23 +397,27 @@ Description:
...
+Setting Name:
 
-The `http3_transport_support` parameter indicates that the specified HTTP/3
-connection is WebTransport-capable.
+: ENABLE_WEBTRANSPORT
 
 Value:
 
-: 0x????
+: 0x???? (currently H2 draft uses 0xFB)
```
DavidSchinazi, line 416: Can we pick a value in the 4-byte encoding range to allow folks to test interop?

draft-vvv-webtransport-http3.md
vasilvv: We should!  Wanna update CONNECT-UDP draft to use stream IDs?

draft-vvv-webtransport-http3.md
```diff
@@ -413,7 +413,7 @@ Setting Name:
 
 Value:
 
-: 0x???? (currently H2 draft uses 0xFB)
+: 0x63e941ac
```
DavidSchinazi, line 416: note that this is the 8-byte encoding not 4-byte :) Also please add the value to the [registry](https://github.com/quicwg/base-drafts/wiki/Temporary-IANA-Registry)

draft-vvv-webtransport-http3.md
DavidSchinazi: No, CONNECT-UDP relies on having separate namespaces for streams and DATAGRAM flow IDs. This feature is used by [the quic-proxy extension](https://tools.ietf.org/html/draft-pauly-masque-quic-proxy-00)

LPardue: I think it would be ok to have the same flow ID vendor work independent of the semantic usage of the connected stream right?

vasilvv: I am not convinced CONNECT-UDP actually relies on this property, at least the draft you link shows no indication on why that is required.

LPardue: @vasilvv I think you're asking a similar question to the one I posed on https://github.com/ietf-wg-masque/draft-ietf-masque-connect-udp/issues/16

DavidSchinazi: @vasilvv if you look at the [Example section](https://tools.ietf.org/html/draft-pauly-masque-quic-proxy-00#section-6) you'll see that multiple CONNECT-UDP requests with this extension can share a flow ID

vasilvv: I read that, I just don't see why they need to share it.

DavidSchinazi: It's about shared lifetimes - it allows multiple requests to come and go and the underlying flow ID to remain valid - otherwise as the older flow IDs get removed you could end up with dropped datagrams. This also allows using smaller encodings for flow ID varints which adds up over time

vasilvv: I'm confused.  You have to keep the stream up for the duration of associated connection ID's lifetime anyways, since once the stream is gone, so are the mappings.

Does the varint thing matter?  I thought the entire point of draft-pauly-masque-quic-proxy is to bring 99% of the traffic outside the DATAGRAMs.

DavidSchinazi: You keep the stream open for its part of the mapping, but you can think of the flow ID as being reference counted by all the streams that use it.

Conceptually, HTTP/3 DATAGRAM Flow IDs have their own namespace similar to how HTTP/3 Push IDs have their own namespace distinct from the Stream ID namespace.

vasilvv: This all sounds like unnecessary complexity to me.  If you don't keep separate flow IDs, you don't need to reference-count them in first place.

afrind:
I was reviewing the H2 WT draft, and I now wonder if "connect stream" is really the best terminology.  Is "session stream" more representative of what it is -- a stream that represents the WebTransport Session?  CONNECT just happens to be the method used to open the stream.

draft-vvv-webtransport-http3.md
```diff
@@ -137,17 +137,17 @@ can be accessed via an HTTP/3 server.
 WebTransport servers in general are identified by a pair of authority value and
 path value (defined in {{!RFC3986}} Sections 3.2 and 3.3 correspondingly).
 
-When an HTTP/3 connection is established, the client and server have to
-negotiate a specific set of QUIC transport parameters that indicate support for
-various WebTransport features.  Most notably, the `http3_transport_support`
-parameter signals WebTransport support to the peer.
+When an HTTP/3 connection is established, both the client and server have to
+negotiate a SETTINGS_ENABLE_WEBTRANSPORT setting in order to indicate that they
```
ianswett, line 141:
```suggestion
send a SETTINGS_ENABLE_WEBTRANSPORT setting in order to indicate that they
```

I'd argue H3 settings aren't really negotiated.

draft-vvv-webtransport-http3.md
ianswett: Let's talk about this on VC sometime, since I'd like to walk through all the pros and cons here.

ianswett (commented):
Thanks Victor, switching to SETTINGs definitely makes sense.  Let's all talk about flow Ids sometime as well.

draft-vvv-webtransport-http3.md
vasilvv: Let's merge this as-is for now, and update when with have consensus in the h3datagram draft.

PR #25, opened 2021-01-22
Add negotiation for pooling support

vasilvv:


draft-vvv-webtransport-http3.md
```diff
@@ -157,10 +157,61 @@ is closed.
...
+* Supported ("1") indicates that the client supports WebTrasnport over HTTP/3
+  and may potentially initiate a WebTransport session in the future.
+* Exclusive ("2") indicates that the client both supports WebTransport over
+  HTTP/3 and has a WebTransport request that it intends to send over the
+  connection in question.
+
+SETTINGS_ENABLE_WEBTRANSPORT sent by the server has the following possible values:
+* Unsupported ("0") indicates that the server does not support WebTransport
+  over HTTP/3.
+* Supported ("1") indicates that the server supports WebTrasnport over HTTP/3.
```
DavidSchinazi, line 182:
```suggestion
* Supported ("1") indicates that the server supports WebTransport over HTTP/3.
```

draft-vvv-webtransport-http3.md
```diff
@@ -157,10 +157,61 @@ is closed.
...
-the parameter has been negotiated.
+MUST send a SETTINGS_ENABLE_WEBTRANSPORT setting.  The value of
+SETTINGS_ENABLE_WEBTRANSPORT is "0" (Unsupported), however it can also assume
+values of "1" (Supported) and "2" (Exclusive).  The exact meaning of the value
+depends on the party.
+
+SETTINGS_ENABLE_WEBTRANSPORT sent by the client has the following possible values:
+* Unsupported ("0") indicates that the client does not support WebTransport
+  over HTTP/3.
+* Supported ("1") indicates that the client supports WebTrasnport over HTTP/3
```
DavidSchinazi, line 173:
```suggestion
* Supported ("1") indicates that the client supports WebTransport over HTTP/3
```

draft-vvv-webtransport-http3.md
```diff
@@ -157,10 +157,61 @@ is closed.
...
 
+In WebTransport, an *exclusive mode* is a state of HTTP/3 connection where the
+connection is expected to have exactly one WebTransport session, and no other
+HTTP requests.
+
 In order to indicate support for WebTransport, both the client and the server
-MUST send a SETTINGS_ENABLE_WEBTRANSPORT value set to "1" in their SETTINGS
-frame.  Endpoints MUST NOT use any WebTransport-related functionality unless
-the parameter has been negotiated.
+MUST send a SETTINGS_ENABLE_WEBTRANSPORT setting.  The value of
```
DavidSchinazi, line 165: We need to state somewhere that if SETTINGS_ENABLE_WEBTRANSPORT is not sent, its value is implicitly set to the default of 0.

draft-vvv-webtransport-http3.md
```diff
@@ -157,10 +157,61 @@ is closed.
...
+  over HTTP/3.
+* Supported ("1") indicates that the server supports WebTrasnport over HTTP/3.
+* Exclusive ("2") indicates that the server expects to handle exactly one
+  request over the connection in question, and that request has to be a
+  WebTransport session.
+
+Once an HTTP/3 implementation has the SETTINGS_ENABLE_WEBTRANSPORT from both
+sides, the following should happen:
+
+* If one of client's or server's value is "0", and another value is either "0"
```
DavidSchinazi, line 190: This doesn't handle the case where client sends 0 and server sends 2. Why not simply say: if any endpoint sent 0 then WebTransport is not supported on this connection - which means clients MUST NOT send the special CONNECT request, and servers MUST fail it with a certain error.

draft-vvv-webtransport-http3.md
```diff
@@ -157,10 +157,61 @@ is closed.
...
+  with error, or it MAY keep it closed for regular HTTP use.
+
+~~~~~~~~~~ drawing
+                     +-------+-------------+----------------+
+                     | C = 0 |    C = 1    |     C = 2      |
+             +-------+-------+-------------+----------------+
+             | S = 0 |  N/A  |     N/A     |      N/A       |
+             +-------+-------+-------------+----------------+
+             | S = 1 |  N/A  | Normal mode |  Normal mode   |
+             +-------+-------+-------------+----------------+
```
DavidSchinazi, line 209: What's the point of having the client tell the server whether it wants to be exclusive or not?

DavidSchinazi (changes requested):
I don't think this proposal works when the server sends its SETTINGS first

LPardue:
A different way to design this would be to say that the setting advertises the maximum number of Http3Transport sessions that the endpoint is willing to support. The maximum number in any session is then the minimum of the client and server value. E.g. if either side sends 0, its off. If either side sends > 0 its on.

DavidSchinazi:
@LPardue your design wouldn't allow a server to say "I'd like either HTTP or WT, but not both on the same connection" though, right?

LPardue:
Sorry, I might have enterd on the wrong axis of thinking about pooling as sharing one connection for WebTransport only, not mixing with conventional HTTP/3. Although, I'm not sure I follow how the proposal here satisfies the ""I'd like either HTTP or WT, but not both on the same connection" either.

DavidSchinazi:
@LPardue In my understanding, this PR doesn't handle the "I'd like either HTTP or WT, but not both on the same connection" use-case. And I think that's what some folks were asking for.

draft-vvv-webtransport-http3.md
vasilvv: No, it handles it with a connection error (third bullet point).

draft-vvv-webtransport-http3.md
vasilvv: It's not about whether it's exclusive or not, it's about whether the client intends to send a WebTransport request or not.

draft-vvv-webtransport-http3.md
vasilvv: Fixed.

draft-vvv-webtransport-http3.md
```diff
@@ -157,10 +157,61 @@ is closed.
...
+HTTP requests.
+
 In order to indicate support for WebTransport, both the client and the server
-MUST send a SETTINGS_ENABLE_WEBTRANSPORT value set to "1" in their SETTINGS
-frame.  Endpoints MUST NOT use any WebTransport-related functionality unless
-the parameter has been negotiated.
+MUST send a SETTINGS_ENABLE_WEBTRANSPORT setting.  The default value of
+SETTINGS_ENABLE_WEBTRANSPORT is "0" (Unsupported), however it can also assume
+values of "1" (Supported) and "2" (Exclusive).  The exact meaning of the value
+depends on the party.
```
DavidSchinazi, line 168:
```suggestion
depends on which endpoint is sending the setting.
```

DavidSchinazi (changes requested):
I still don't understand how this works. How does the server know what value to send?

draft-vvv-webtransport-http3.md
```diff
@@ -157,10 +157,61 @@ is closed.
...
+
+SETTINGS_ENABLE_WEBTRANSPORT sent by the client has the following possible values:
+* Unsupported ("0") indicates that the client does not support WebTransport
+  over HTTP/3.
+* Supported ("1") indicates that the client supports WebTransport over HTTP/3
+  and may potentially initiate a WebTransport session in the future.
+* Exclusive ("2") indicates that the client both supports WebTransport over
+  HTTP/3 and has a WebTransport request that it intends to send over the
+  connection in question.
```
afrind, line 177: Does this mean the client has to have a request in mind before sending the SETTING?  Is pre-connecting allowed?

draft-vvv-webtransport-http3.md
```diff
@@ -157,10 +157,61 @@ is closed.
...
+
+Once an HTTP/3 implementation has the SETTINGS_ENABLE_WEBTRANSPORT from both
+sides, the following should happen:
+
+* If one of client's or server's value is "0", and another value is either "0"
+  or "1", the connection does not support WebTransport.
+* If both server's value is "1", and client's value is "1" or "2", the
+  connection supports WebTransport in normal mode.
+* If server's value is "2", and client's value is "0" or "1", the connection
+  MUST be closed with a connection error.
```
afrind, line 195: Echoing @DavidSchinazi's earlier point -- this makes it challenging to run a web server and an exclusive-only web transport server on the same address/port.  The server can send SETTINGS in 0.5-RTT, and it won't know if the client is going to send 0 (HTTP) or 2 (WT).  Alternately, the server can wait to receive the client SETTINGS before sending its own, but that's suboptimal for other reasons.

afrind: Somewhat related is that there is a completely incompatible set of modes here -- C=1, S=2.  These clients can't talk at all, even if the client would prefer pooled connections but would settle for exclusive ones.

draft-vvv-webtransport-http3.md
vasilvv: Sure, as long as you're fine with the resulting connection being usable for a single WebTransport.

draft-vvv-webtransport-http3.md
vasilvv: I'm confused.  I don't think you can run a non-exclusive server and and exclusive server on the same port without multiplexing by something like ALPN.

If the client prefers pooled version, it can send C=1.  C=1, S=1 results in pooled WebTransport, C=1, S=2 results in exclusive WebTransport.

vasilvv:
It sounds like calling C=2 "exclusive" was a mistake.  Renamed it.

draft-vvv-webtransport-http3.md
afrind: Per the text and the table, C=1, S=2 is an error.

vasilvv:
Abandoning per discussion on #34.

Issue #23, opened 2020-11-16
Http3Transport server-initiated bidirectional stream future proofing


kixelated:
Http3Transport currently assumes that server-initiated bidirectional streams and datagrams can only be used for WebTransport. 

I suggest we prefix each server-initiated bidirectional stream and datagram with a new HTTP/3 frame type for WebTransport. This way, a future standard could define it's own frame type to run alongside WebTransport.

ianswett:
Using an H3 frame type for server initiated bidi streams would also make the design more consistent, instead of the asymmetric design there is today.  It costs a byte per bidi stream, but it seems worth it for consistency and flexibility.

For datagrams, I believe that we'd want to be able to coexist with other H3 extensions, such as MASQUE, and I believe that's possible without any extra overhead by adding a Session -> Datagram flow ID mapping when a WebTransport session is created.

afrind:
If it's a single byte, it's more of a stream type (like the unidirectional stream-type) than a frame type, right?

> For datagrams, I believe that we'd want to be able to coexist with other H3 extensions, such as MASQUE, and I believe that's possible without any extra overhead by adding a Session -> Datagram flow ID mapping when a WebTransport session is created.

There's a discussion Victor pointed me to in #21, where it's suggested that datagrams belonging to a WebTransport session use the Flow ID = to the Session ID.  It would be nice to avoid the extra state and lookup.

ianswett:
Ah yes, it's more of a stream type.  In that case, I'm unsure of whether there's a good way of eliminating the asymmetry.  The only option I can think of is to not use a stream type for server bidi streams and only use the frame, like for the client initiated bidi streams?

Thanks for the pointer to #21, it looks like there are a few potential designs for datagrams flow IDs.  If in doubt, simpler is probably better.  I'll think more about this one.

afrind:
I agree we should use the same mechanics for client and server bidi streams.

ianswett:
After thinking about this more, not having a stream type doesn't seem so bad to me and it does provide symmetry.

Stream types seem like an obvious design choice, but I believe HTTP/3 is sufficiently constrained that one could infer the stream type based on the first frame on the stream.  For example, the control stream MUST start with SETTINGS.  In retrospect, having both stream types and frame types is a bit awkward, and it likely would have been cleaner to only have frame types, but that ship has sailed.

afrind:
[...]

PR #46, opened 2021-05-11
Redefine server-initiated bidi streams

vasilvv:
This matches agreement at the last interim.  We should separate this
into a different draft, but for now this will make the draft match
implementation.

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -280,11 +280,11 @@ HTTP/3 does not by itself define any semantics for server-initiated
 bidirectional streams.  If WebTransport setting is negotiated by both
 endpoints, the syntax of the server-initiated bidirectional streams SHALL be
 the same as the syntax of client-initated bidirectional streams, that is, a
-sequence of HTTP/3 frames.  The only allowed frame is WEBTRANSPORT_STREAM;
-subsequent documents may add more frames, or define server-initiated semantics
-for existing frames.
+sequence of HTTP/3 frames.  The only frame defined by this document for use
+within server-initiated bidirectional streams is WEBTRANSPORT_STREAM.
 
-TODO: move the paragraph above into a separate draft.
+TODO: move the paragraph above into a separate draft; define what happens with
+already existing HTTP/3 frames on server-initiated bidirectional streams.
 
 ## Datagrams
 
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -254,9 +254,9 @@ by the user-specified stream data ({{fig-unidi}}).
 ~~~~~~~~~~
 {: #fig-unidi title="Unidirectional WebTransport stream format"}
 
-## Client-Initiated Bidirectional Streams
+## Bidirectional Streams
 
-WebTransport clients can initiate bidirectional streams by opening an HTTP/3
+WebTransport endpoints can initiate bidirectional streams by opening an HTTP/3
 bidirectional stream and sending an HTTP/3 frame with type
 `WEBTRANSPORT_STREAM` (type=0x41).  The format of the frame SHALL be the frame
 type, followed by the session ID, encoded as a variable-length integer,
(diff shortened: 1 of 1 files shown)


vasilvv:
Partially addresses #23.

cc @kixelated @afrind

draft-ietf-webtrans-http3.md
```diff
@@ -276,27 +276,15 @@ SHALL last until the end of the stream.
...
-  0                   1                   2                   3
-  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
-  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-  |                        Session ID (i)                       ...
-  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-  |                         Stream Body                         ...
-  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-~~~~~~~~~~
-{: #fig-bidi-server title="Server-initiated bidirectional stream format"}
+TODO: move the paragraph above into a separate draft.
```
afrind, line 287: I think maybe all you need to say is 

```
The only allowed HTTP/3 frame on a WebTransport bidirectional stream is WEBTRANSPORT_STREAM.
```

I don't see a lot of value here saying that some other document might specify something else.

draft-ietf-webtrans-http3.md
```diff
@@ -276,27 +276,15 @@ SHALL last until the end of the stream.
 ~~~~~~~~~~
 {: #fig-bidi-client title="WEBTRANSPORT_STREAM frame format"}
 
-## Server-Initiated Bidirectional Streams
+HTTP/3 does not by itself define any semantics for server-initiated
+bidirectional streams.  If WebTransport setting is negotiated by both
+endpoints, the syntax of the server-initiated bidirectional streams SHALL be
+the same as the syntax of client-initated bidirectional streams, that is, a
+sequence of HTTP/3 frames.  The only allowed frame is WEBTRANSPORT_STREAM;
```
DavidSchinazi, line 283: I would suggest instead allowing all HTTP/3 frame types, and saying that clients MUST silently ignore all frame types except for WEBTRANSPORT_STREAM. That'll enable future extensibility without requiring a SETTING. This will also make it easier to split this into its own draft.

afrind: But this frame is a TV frame (not TLV), so it's defined to take up the entire stream?  

To get more extensibility, we could have server-initiated bidi stream carry a 'Stream Type', like uni streams, followed by a series of HTTP/3 frames.  WebTransport would then define its TV frame that converts the stream to unframed, and other types of server-initiated bidi streams could use TLV or TV frames.

DavidSchinazi: The frame being TV means that it needs to be the *last* frame on a stream. It doesn't need to be the *only* frame on a stream. All I'm saying is that server-initiated bidi streams can carry any and all h3 frames, and when they see a WEBTRANSPORT_STREAM frame then the rest of the stream is taken over. Just like client-initiated bidi streams.

afrind: The frame is currently defined to be both first (sets the type to WT) and last (it's TV).  

I'm suggesting you might want to split those two functions, and all streams (except client-initiated bidi) start with a stream type, followed by frames.

LPardue: I don't think we want to allow _all_ frame types. In particular, allowing HEADERS to be sent and but dropped at the receiver seems like a good way to break QPACK (see https://datatracker.ietf.org/doc/html/draft-ietf-quic-qpack-21#section-2.2.2.1). Opening things up for other frames types, especially new extension frames, seems ok though.

DavidSchinazi: @afrind what does "sets the type" mean? bidi streams don't have a type. It makes sense to allow extension frames before jumping into WT.

@LPardue that's a good point, maybe not all frames, but definitely extensions

afrind: @DavidSchinazi : This draft defines a frame with two fields - one is the frame type, which serves multiple functions, and the session ID.  As written, sending this frame identifies the stream as a WT stream (semantics governed by this draft), associates this stream with a WT session, and converts the stream to unframed.

For general server bidi-stream extension semantics, I prefer we send a single 'type' field (not a frame type) to identify the semantics of the stream, followed by frames.  Then this draft would be slightly different:

```
[ Stream Type=WT Server-Init Bidi Stream ] [ Frame Type=WT_STREAM | Session ID ]
```

@LPardue : I don't think we should disallow any QPACK bearing frames in extensions, any more than H2 forbid HPACK bearing frames.  You just have to ensure that the peer has expressed support for the extension before using the dynamic table.  The static table is always OK.

LPardue: oh I'm not trying to prevent new things that can carry QPACK, I just think you'll want to couple that with a negotiation that makes it explicit the other end would respect the rules. Discarding HEADERS seems like an odd choice of behaviour to enable.

vasilvv: It sounds like this question needs further discussion.  I am leaving the frame-based version here to get us to a draft we can interop on, but we can continue this discussion on issue #23 (and at the upcoming interim).

draft-ietf-webtrans-http3.md
vasilvv: Made text more vague; let's discuss this further on #23 and at the upcoming interim.

DavidSchinazi (approved):

PR #59, opened 2021-07-29
Allocate HTTP/3 error codes for use by the WebTransport application

### LINKED ISSUE #40: RESET_STREAM error codes
If we are going to let the API users supply their own stream reset code, we should figure out how this is represented on the wire.  We could potentially use the error codes as-is and say they mean different things for WebTransport streams than what they mean for normal streams, but I am not sure this is sound as we don't always know that a stream is a WebTransport stream at the time we process the reset (also this sounds like a pain to implement).

One possibility would be to reserve a block within HTTP/3 error code space.  However, since HTTP/3 error codes contain GREASE code-points throughout, we might end up having to "de-grease" and "re-grease" them as the largest continuous block only has 255 codepoints available.

Alternatively, if we decide to just send RESET_STREAM as a part of CONNECT stream due to issues in #31, we could just send the error code there, and keep one single code for user-initiated termination.

Whatever we decide here should probably work for H/2 as well.

cc @martinthomson

vasilvv:
Fixes #40.

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -286,7 +286,7 @@ codes.  An example pseudocode can be seen in {{fig-remap-errors}}.
     last = 0x52e4a40fa9e2
 
     def webtransport_code_to_http_code(n):
-        return first + n + n // 0x1e
+        return first + n + floor(n / 0x1e)
 
     def http_code_to_webtransport_code(h):
         assert(first <= h <= last)
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -265,6 +265,54 @@ within server-initiated bidirectional streams is WEBTRANSPORT_STREAM.
 TODO: move the paragraph above into a separate draft; define what happens with
 already existing HTTP/3 frames on server-initiated bidirectional streams.
 
+## Resetting Data Streams
+
+A WebTransport endpoint may send a RESET_STREAM or a STOP_SENDING frame for a
+WebTransport data stream.  Those signals are propagated by the WebTransport
+implementation to the application.
+
+A WebTransport application SHALL provide an error code for those operations.
+Since WebTransport shares the error code space with HTTP/3, WebTransport
+application errors for streams are limited to an unsigned 8-bit integer,
+assuming values between 0x00 and 0xff.  WebTransport implementations SHALL
+remap those error codes into an error range where 0x00 corresponds to
+0x52e4a40fa8db, and 0xff corresponds to 0x52e4a40fa9e2.  Note that there are
+code points inside that range of form "0x1f * N + 0x21" that are reserved by
+{{Section 8.1 of HTTP3}}; those have to be accounted for when mapping the error
+codes.  An example pseudocode can be seen in {{fig-remap-errors}}.
+
+~~~~~~~~~~
+    first = 0x52e4a40fa8db
+    last = 0x52e4a40fa9e2
+
+    def webtransport_code_to_http_code(n):
+        return first + n + n // 0x1e
+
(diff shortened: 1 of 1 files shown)


vasilvv:
cc @martinthomson

draft-ietf-webtrans-http3.md
```diff
@@ -265,6 +265,54 @@ within server-initiated bidirectional streams is WEBTRANSPORT_STREAM.
...
+code points inside that range of form "0x1f * N + 0x21" that are reserved by
+{{Section 8.1 of HTTP3}}; those have to be accounted for when mapping the error
+codes.  An example pseudocode can be seen in {{fig-remap-errors}}.
+
+~~~~~~~~~~
+    first = 0x52e4a40fa8db
+    last = 0x52e4a40fa9e2
+
+    def webtransport_code_to_http_code(n):
+        return first + n + n // 0x1e
```
DavidSchinazi, line 289: How about avoiding `//` syntax that not everyone knows:
```suggestion
        return first + n + floor(n / 0x1e)
```

DavidSchinazi (approved):

draft-ietf-webtrans-http3.md
```diff
@@ -265,6 +265,54 @@ within server-initiated bidirectional streams is WEBTRANSPORT_STREAM.
...
+implementation to the application.
+
+A WebTransport application SHALL provide an error code for those operations.
+Since WebTransport shares the error code space with HTTP/3, WebTransport
+application errors for streams are limited to an unsigned 8-bit integer,
+assuming values between 0x00 and 0xff.  WebTransport implementations SHALL
+remap those error codes into an error range where 0x00 corresponds to
+0x52e4a40fa8db, and 0xff corresponds to 0x52e4a40fa9e2.  Note that there are
+code points inside that range of form "0x1f * N + 0x21" that are reserved by
+{{Section 8.1 of HTTP3}}; those have to be accounted for when mapping the error
```
ekinnear, line 281: Great to have pseudocode, is there a chance we could explain _how_ someone is expected to account for those (i.e. do you just consider those invalid/taken already or shift by that much each time you pass an already reserved one) in text so that when people get to the pseudocode they know what behavior they're supposed to be seeing?

draft-ietf-webtrans-http3.md
```diff
@@ -265,6 +265,54 @@ within server-initiated bidirectional streams is WEBTRANSPORT_STREAM.
...
+{: #fig-remap-errors title="Pseudocode for converting between WebTransport
+application errors and HTTP/3 error codes; here, `//` is integer division" }
+
+WebTransport data streams are associated with sessions through a header at the
+beginning of the stream; resetting a stream may result in that data being
+discarded.  Because of that, WebTransport application error codes are best
+effort, as the WebTransport stack is not always capable of associating the
+reset code with a session.  The only exception is the situation where there is
+only one session on a given HTTP/3 connection, and no intermediaries between
+the client and the server.
```
ekinnear, line 306: Is the client (or the server) expected to be aware of the presence of intermediaries for this?

draft-ietf-webtrans-http3.md
vasilvv: You account by shifting.  Clarified that.

draft-ietf-webtrans-http3.md
vasilvv: I think the way I'd put it is that the web application (i.e. the combination of client and server) has to be aware of lack of intermediaries to rely on those.

DavidSchinazi: Having the application be aware of intermediaries sounds like a brittle design, as the presence of intermediaries might be out of the control of the application developer

Issue #22, opened 2020-11-16
Http3Transport connection flow control


kixelated:
Note that all of the below works when you swap "server" and "client". I wanted to focus on a poor Javascript client implementation, but it's equally possible to run into these cases with a poor server implementation.

Let's say that you have a Http3Transport session that shares a connection with HTTP/3 or another Http3Transport session. 

### MAX_DATA
The client has limited buffer space and hypothetically advertises MAX_DATA of 1MB for the connection. If the server pushes 1MB of unprocessed data over a Http3Transport session, then it will prevent the server from sending data over the connection, including HTTP/3 responses and other Http3Transport sessions.

This can occur if the application does not accept all remote-created streams or does not fully read every stream. This will be VERY easy for a developer to mess up. It can also occur if the application can not process data faster than it is received.

In my case, the server creates unidirectional streams and pushes a large amount of media over them. The flow control limit would be hit if the Javascript application:

- Only accepts bidirectional streams (ex. initialization)
- Does not accept a unidirectional stream (ex. error or teardown)
- Does not fully read or abort every stream (ex. parsing error)
- Is not able to parse the media fast enough (ex. underpowered).

This is mostly intended behavior, otherwise these "bugs" would cause unbounded memory usage. However, the blast radius is bigger when sharing a connection. 

In my use-case, hitting the flow control limit would also block any HTTP/3 RPC calls or unrelated traffic on the connection. This introduces more scenarios for the connection to become deadlocked or generally stalled.

### MAX_STREAMS
HTTP/3 implementations use MAX_STREAMS to limit the number of simultaneous requests. This use-case no longer works when sharing a connection with a Http3Transport session, and HTTP/3 implementations must instead return explicit error status codes when this threshold is reached.

The same issues with MAX_DATA also apply to MAX_STREAMS. A Javascript application that fails to accept or read/abort every stream will consume flow control resources. This can manifest itself in strange ways, such as HTTP/3 being unable to create control streams, or simply being unable to create new request streams.

LPardue:
Thanks for the salient overview, it makes it easier to understand your concerns.

[...]

PR #74, opened 2022-06-27
Add SETTINGS_MAX_WEBTRANSPORT_SESSIONS settings parameter

### LINKED ISSUE #34: Negotiation of pooling support
The H3 draft by its nature allows for the parallel processing of multiple WT sessions and also the intermixing of normal H3 request with WT.  Server endpoints may wish to negotiate or disable this behavior.  Clients can simply choose not to do either kind of pooling if they don't want to.

There's already a PR open (#25) with a design idea.

vasilvv:
Also bump the protocol wire vesrion

Fixes #34

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -147,6 +147,12 @@ MUST NOT process any incoming WebTransport requests until the client settings
 have been received, as the client may be using a version of WebTransport
 extension that is different from the one used by the server.
 
+In addition to the setting above, the server MUST send a
+SETTINGS_MAX_WEBTRANSPORT_SESSIONS parameter indicating the maximum number of
+concurrent sessions it is willing to receive.  The default value for the
+SETTINGS_MAX_WEBTRANSPORT_SESSIONS parameter is "0", meaning that the server is
+not willing to receive any WebTransport sessions.
+
 ## Extended CONNECT in HTTP/3
 
 {{!RFC8441}} defines an extended CONNECT method in Section 4, enabled by the
@@ -187,18 +193,15 @@ The `webtransport` HTTP Upgrade Token uses the Capsule Protocol as defined in
 
 ## Limiting the Number of Simultaneous Sessions
 
-From the flow control perspective, WebTransport sessions count against the
-stream flow control just like regular HTTP requests, since they are established
-via an HTTP CONNECT request.  This document does not make any effort to
-introduce a separate flow control mechanism for sessions, nor to separate HTTP
-requests from WebTransport data streams.  If the server needs to limit the rate
-of incoming requests, it has alternative mechanisms at its disposal:
-
-* `HTTP_REQUEST_REJECTED` error code defined in [HTTP3] indicates to the
-  receiving HTTP/3 stack that the request was not processed in any way.
-* HTTP status code 429 indicates that the request was rejected due to rate
-  limiting {{!RFC6585}}.  Unlike the previous method, this signal is directly
-  propagated to the application.
+This document defines a SETTINGS_MAX_WEBTRANSPORT_SESSIONS parameter that
+allows the server to limit the maximum number of concurrent WebTransport
+sessions on a single HTTP/3 connection.  The client MUST NOT open more sessions
+than indicated in the server SETTINGS parameters.  The server MUST NOT close
+the connection if the client opens sessions exceeding this limit, as the client
+and the server do not have a consistent view of how many streams are open due
+to the asynchronous nature of the protocol; instead, it MUST reset all of the
+CONNECT streams it is not willing to process with the `HTTP_REQUEST_REJECTED`
+status defined in [HTTP3].
 
 # WebTransport Features
 
@@ -471,7 +474,7 @@ Setting Name:
 
 Value:
 
-: 0x2b603742
+: 0x8b422342
 
 Default:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-webtrans-http3.md
```diff
@@ -187,18 +193,15 @@ The `webtransport` HTTP Upgrade Token uses the Capsule Protocol as defined in
...
-  receiving HTTP/3 stack that the request was not processed in any way.
-* HTTP status code 429 indicates that the request was rejected due to rate
-  limiting {{!RFC6585}}.  Unlike the previous method, this signal is directly
-  propagated to the application.
+This document defines a SETTINGS_MAX_WEBTRANSPORT_SESSIONS parameter that
+allows the server to limit the maximum number of concurrent WebTransport
+sessions on a single HTTP/3 connection.  The client MUST NOT open more sessions
+than indicated in the server SETTINGS parameters.  The server MUST NOT close
+the connection if the client opens sessions exceeding this limit, as the client
+and the server do not have a consistent view of how many streams are open due
```
marten-seemann, line 201:
```suggestion
and the server do not have a consistent view of how many sessions are open due
```

draft-ietf-webtrans-http3.md
```diff
@@ -147,6 +147,12 @@ MUST NOT process any incoming WebTransport requests until the client settings
 have been received, as the client may be using a version of WebTransport
 extension that is different from the one used by the server.
 
+In addition to the setting above, the server MUST send a
+SETTINGS_MAX_WEBTRANSPORT_SESSIONS parameter indicating the maximum number of
```
DavidSchinazi, line 151: I thought we were going to only have one setting since SETTINGS_MAX_WEBTRANSPORT_SESSIONS=0 means the same as "WebTransport not supported".

DavidSchinazi:
Additionally, this PR addresses #34 but not #22 so it shouldn't be marked as fixing #22.

draft-ietf-webtrans-http3.md
vasilvv: We can't do that at this point, since only the server sends SETTINGS_MAX_WEBTRANSPORT_SESSIONS, and we need a setting from the client to do version negotiation.

vasilvv:
> Additionally, this PR addresses #34 but not #22 so it shouldn't be marked as fixing #22.

I thought we've agreed at some point that this is all we want in terms of solving #22?

draft-ietf-webtrans-http3.md
DavidSchinazi: Can't we use only the header for VN and drop the setting? If anyone is implementing draft versions with an intermediary they can parse the headers.

DavidSchinazi:
> > Additionally, this PR addresses #34 but not #22 so it shouldn't be marked as fixing #22.
> 
> I thought we've agreed at some point that this is all we want in terms of solving #22?

I don't recall that (but I could be misremembering), do you have a reference to point me to?

draft-ietf-webtrans-http3.md
vasilvv: > > > Additionally, this PR addresses #34 but not #22 so it shouldn't be marked as fixing #22.
> > 
> > 
> > I thought we've agreed at some point that this is all we want in terms of solving #22?
> 
> I don't recall that (but I could be misremembering), do you have a reference to point me to?

I checked the meeting notes for the design team, and it looks like we decided the opposite of what I remember, so it looks like I'm the one misremembering here.

vasilvv: From what I remember from the way our code works, that would work poorly, so I'm hesitant to do that.

ekinnear (approved):
Seems good, will also pull into the design team branch and also for H2 as well

draft-ietf-webtrans-http3.md
ekinnear: This is written such that MAX_WEBTRANSPORT_SESSIONS is only sent by the server, do we want to have enable_webtransport (or version) be still sent by both/the client, or do we think that using the upgrade token indicates support on the part of the client? (And is that easy enough to handle in code)

draft-ietf-webtrans-http3.md
```diff
@@ -187,18 +193,15 @@ The `webtransport` HTTP Upgrade Token uses the Capsule Protocol as defined in
...
-  propagated to the application.
+This document defines a SETTINGS_MAX_WEBTRANSPORT_SESSIONS parameter that
+allows the server to limit the maximum number of concurrent WebTransport
+sessions on a single HTTP/3 connection.  The client MUST NOT open more sessions
+than indicated in the server SETTINGS parameters.  The server MUST NOT close
+the connection if the client opens sessions exceeding this limit, as the client
+and the server do not have a consistent view of how many sessions are open due
+to the asynchronous nature of the protocol; instead, it MUST reset all of the
+CONNECT streams it is not willing to process with the `HTTP_REQUEST_REJECTED`
+status defined in [HTTP3].
```
ekinnear, line 204: (Just to disambiguate "status")
```suggestion
error code defined in [HTTP3].
```

draft-ietf-webtrans-http3.md
```diff
@@ -187,18 +193,15 @@ The `webtransport` HTTP Upgrade Token uses the Capsule Protocol as defined in
...
-  limiting {{!RFC6585}}.  Unlike the previous method, this signal is directly
-  propagated to the application.
+This document defines a SETTINGS_MAX_WEBTRANSPORT_SESSIONS parameter that
+allows the server to limit the maximum number of concurrent WebTransport
+sessions on a single HTTP/3 connection.  The client MUST NOT open more sessions
+than indicated in the server SETTINGS parameters.  The server MUST NOT close
+the connection if the client opens sessions exceeding this limit, as the client
+and the server do not have a consistent view of how many sessions are open due
+to the asynchronous nature of the protocol; instead, it MUST reset all of the
+CONNECT streams it is not willing to process with the `HTTP_REQUEST_REJECTED`
```
LPardue, line 203: it's formally this

```suggestion
CONNECT streams it is not willing to process with the `H3_REQUEST_REJECTED`
```

vasilvv:
Superseded by #86.

PR #75, opened 2022-06-27
Add a reset error code for closing streams due to session closure

vasilvv:


draft-ietf-webtrans-http3.md
```diff
@@ -350,8 +350,9 @@ following conditions is met:
 * a CLOSE_WEBTRANSPORT_SESSION capsule is either sent or received.
 
 Upon learning that the session has been terminated, the endpoint MUST reset all
-of the streams associated with the session; it MUST NOT send any new datagrams
-or open any new streams.
+of the streams associated with the session with the
```
DavidSchinazi, line 353: note that reset only affects the write side of the stream, we also need to abort reading. Here's example text from RFC 9114:


> Implementations SHOULD cancel requests by abruptly terminating any directions of a stream that are still open. To do so, an implementation resets the sending parts of streams and aborts reading on the receiving parts of streams; see [Section 2.4](https://www.rfc-editor.org/rfc/rfc9000#section-2.4) of RFC 9000.

DavidSchinazi (approved):

ekinnear (approved):

draft-ietf-webtrans-http3.md
```diff
@@ -546,6 +548,23 @@ Specification:
...
+
+: H3_WEBTRANSPORT_SESSION_GONE
+
+Value:
+
+: 0x170d7b68
+
+Description:
+
+: WebTransport data stream reset because the associated WebTransport session
```
LPardue, line 561: since the text applies to both read and write, we can be more generic in the description

```suggestion
: WebTransport data stream aborted because the associated WebTransport session
```

PR #79, opened 2022-07-22
Add a capsule to drain a WebTransport session

### LINKED ISSUE #27: Draining WebTransport Sessions
An HTTP/3 server needs the ability to gracefully drain existing connections. For HTTP/3, this involves sending a GOAWAY frame indicating that future requests should use a new HTTP/3 connection. However, the behavior of GOAWAY is partially undefined with WebTransport sessions. 

After receiving a GOAWAY, the client is no longer able to make new CONNECT requests. This means that any new WebTransport sessions will need to use a new HTTP/3 connection. This is great.

However, it's not clear how existing sessions should behave after receiving a GOAWAY frame. Are they prohibited from creating new streams? How does this interact with the stream ID sent in the GOAWAY frame? Should the application expose the receipt of a GOAWAY to start a shutdown process?

afrind:
Fixes #27

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -329,6 +329,20 @@ requests for new WebTransport sessions.  A GOAWAY frame does not affect data
 streams for existing WebTransport sessions; those can continue to be opened
 even after the GOAWAY frame has been sent or received.
 
+To drain a WebTransport session, either endpoint can send a
+DRAIN_WEBTRANSPORT_SESSION capsule.  An endpoint SHOULD NOT initiate new
+WebTransport streams or datagram flows after sending or receiving a
+DRAIN_WEBTRANSPORT_SESSION capsule.  Both endpoints can continue sending data on
+existing streams and datagram flows, but SHOULD attempt to gracefully terminate
+the session as quickly as possible.
+
+~~~
+DRAIN_WEBTRANSPORT_SESSION Capsule {
+  Type (i) = DRAIN_WEBTRANSPORT_SESSION,
+  Length (i) = 0
+}
+~~~
+
 # Session Termination
 
 A WebTransport session over HTTP/3 is considered terminated when either of the
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -330,11 +330,9 @@ streams for existing WebTransport sessions; those can continue to be opened
 even after the GOAWAY frame has been sent or received.
 
 To drain a WebTransport session, either endpoint can send a
-DRAIN_WEBTRANSPORT_SESSION capsule.  An endpoint SHOULD NOT initiate new
-WebTransport streams or datagram flows after sending or receiving a
-DRAIN_WEBTRANSPORT_SESSION capsule.  Both endpoints can continue sending data on
-existing streams and datagram flows, but SHOULD attempt to gracefully terminate
-the session as quickly as possible.
+DRAIN_WEBTRANSPORT_SESSION capsule.  After sending or receiving a
+DRAIN_WEBTRANSPORT_SESSION capsule, an endpoint MAY continue using the session
+but SHOULD attempt to gracefully terminate the session as soon as possible.
 
 ~~~
 DRAIN_WEBTRANSPORT_SESSION Capsule {


draft-ietf-webtrans-http3.md
```diff
@@ -329,6 +329,20 @@ requests for new WebTransport sessions.  A GOAWAY frame does not affect data
 streams for existing WebTransport sessions; those can continue to be opened
 even after the GOAWAY frame has been sent or received.
 
+To drain a WebTransport session, either endpoint can send a
+DRAIN_WEBTRANSPORT_SESSION capsule.  An endpoint SHOULD NOT initiate new
+WebTransport streams or datagram flows after sending or receiving a
```
DavidSchinazi, line 334: "Datagram flows" is not a well-defined term. Is the intent to discourage all datagrams?

DavidSchinazi (approved):
Overall looks good modulo one comment

draft-ietf-webtrans-http3.md
```diff
@@ -329,6 +329,20 @@ requests for new WebTransport sessions.  A GOAWAY frame does not affect data
...
+DRAIN_WEBTRANSPORT_SESSION capsule.  Both endpoints can continue sending data on
```
ekinnear, line 335: Perhaps the same note we have about max sessions that the endpoints don't agree on how many there are so please don't try to enforce the limit? Or even just reference that?

ekinnear (approved):

draft-ietf-webtrans-http3.md
afrind: I agree it's not well defined.  The intent was to prevent the application from starting new work over this session, but to allow it to finish existing work.  I don't think we should prevent sending all datagrams -- @DavidSchinazi do you have any text suggestions?

draft-ietf-webtrans-http3.md
afrind: @ekinnear : I'm not following how the stream limit is related here?

draft-ietf-webtrans-http3.md
DavidSchinazi: Maybe use the text about existing work vs new work then?

vasilvv: I'm not actually sure we should give specific guidance here.  "Stop opening new streams" makes sense for HTTP/3, but different protocols may have different drain sequence.  We should probably just describe what it means.

afrind: Maybe something simple like:

```
To drain a WebTransport session, either endpoint can send a DRAIN_WEBTRANSPORT_SESSION capsule.  
After sending or receiving a DRAIN_WEBTRANSPORT_SESSION capsule, an endpoint MAY continue using the session but SHOULD attempt to gracefully terminate the session as soon as possible.
```

aboba:
FYI, the API PR for receiving a DRAIN_WEBTRANSPORT_SESSION capsule capsule is here: https://github.com/w3c/webtransport/pull/437

After receiving a DRAIN_WEBTRANSPORT_SESSION capsule from the server, the client can no longer open unidirectional or bidirectional streams, but I assumed that continuing to send datagrams was ok.  Is that correct?

draft-ietf-webtrans-http3.md
ekinnear: Sorry, 
> An endpoint SHOULD NOT initiate new WebTransport streams or datagram flows after sending or receiving a DRAIN_WEBTRANSPORT_SESSION capsule.

Do both sides agree on the ordering between a new stream coming in vs. a DRAIN_WEBTRANSPORT_SESSION capsule? If not, then we may want to note that so that people plan on having them happen in any order.

draft-ietf-webtrans-http3.md
afrind: Comment from editors -- replace endpoint with client to make it clear there is no action from the user-agent.

aboba:
Section 4.6 of WebTransport over HTTP/3 (https://www.ietf.org/archive/id/draft-ietf-webtrans-http3-04.html#name-interaction-with-http-3-goa) now covers the HTTP/3 GOAWAY frame. When sent from the server -> client this indicates that client attempts to construct additional WebTransport objects will fail. 

Does receipt of a GOAWAY frame need to be indicated to the application, or can it just deal with a (future) constructor error? 

The DRAIN_WEBTRANSPORT_SESSION capsule PR has been merged, but the draft including the merged PR has not been published yet.

PR #80, opened 2022-07-24
clarify when the client can open streams and send datagrams

### LINKED ISSUE #73: opening WebTransport streams before the WebTransport session is established
In order to save a round-trip on connection establishment, a client could optimistically open streams just after sending the CONNECT request, before it has received a reply from the server.

As both unidirectional and bidirectional streams are identified by the session ID, the server will be able to associate these streams with the WebTransport session as soon as it has processed the CONNECT request.

In case the server rejects the CONNECT request, the client should probably reset / cancel these streams. Note that this doesn't require any additional logic on the server side, as the server needs to be prepared to receive streams for sessions that it didn't establish anyway (a malicious client might open streams with arbitrary session IDs, and a server should better be prepared for that).

Currently, the spec forbids opening unidirectional streams before the session is established:
> Once established, both endpoints can open unidirectional streams.

There's no such restriction for bidirectional stream.

marten-seemann:
Fixes #73.

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -189,13 +189,19 @@ Session IDs are used to demultiplex streams and datagrams belonging to different
 WebTransport sessions.  On the wire, session IDs are encoded using the QUIC
 variable length integer scheme described in {{!RFC9000}}.
 
+The client MAY optimistically open unidirectional and bidirectional streams, as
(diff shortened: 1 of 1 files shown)


DavidSchinazi (approved):
(no hats)

ekinnear (approved):

afrind (approved):

Issue #85, opened 2022-10-29
MAX_STREAMS and flow control


ekinnear:
From https://github.com/ietf-wg-webtrans/draft-ietf-webtrans-http3/issues/22#issuecomment-1077698720:
> figure out MAX_STREAMS, a way to limit the number of streams within a session

We've addressed the rest of the plan except for this part via the capsule DT, but there are still open questions about reordering of closed streams and their interaction with stream flow control limits via MAX_STREAMS.

ekinnear:
If reliable resets happen in QUIC, then it should be possible to model this after how QUIC MAX_STREAMS works, and we can just look at the biggest numbered stream.

DavidSchinazi:
Chair: discussed at IETF 116. Consensus in room was to punt this unless someone proposes a solution. More specifically, we'll keep this issue open until IETF 117. If someone proposes a PR (or separate draft) that adds flow control, we will give them agenda time at IETF 117 and discuss whether to include it based on the details of the proposal. If no one has proposed anything by then, we will close the issue with no action.

DavidSchinazi:
Chair: discussed in editor's meeting. @ekinnear and @martinthomson wrote up a proposal. Action Item is with @vasilvv to review.

martinthomson:
https://www.ietf.org/archive/id/draft-thomson-webtrans-session-limit-00.html is the proposal.  We might also want to consider the inclusion of the WebTransport-Init header field to govern initial values for those.  And then, see https://github.com/ietf-wg-webtrans/draft-ietf-webtrans-http2/pull/78 regarding settings to additionally govern these.

DavidSchinazi:
Discussed at editor's meeting: plan is to discuss at 117. @ekinnear can you make slides please?

DavidSchinazi:
Chair: discussed at 117. Lots of discussion in the room, but consensus in room on the following:
* we can't do pooling without flow control as that would cause starvation
* we want JavaScript to not get measurable changes when the session is pooled vs not pooled
* therefore, we can't disable flow control when pooling is disabled
* this is hard to implement
* implementers are focusing on other parts of WebTransport first
* for now, we'll wait on more implementer experience
* we will not publish the document without flow control

ekinnear:
Discussed at length in editor's meeting. @ekinnear to file an issue with the W3C. Needs more discussion here, too.

vasilvv:
[...]

PR #86, opened 2022-10-29
Capsule design team output [HTTP/3]

### LINKED ISSUE #22: Http3Transport connection flow control

### LINKED ISSUE #34: Negotiation of pooling support

### LINKED ISSUE #66: Folding datagrams/streams into the CONNECT stream
Should the protocol share framing semantics with the HTTP/2 draft so that it is easy/easier to take datagrams or streams and put them on the CONNECT stream?

Should it be possible to use that mechanism, or should it instead be forbidden?

As much as possible, it would be desirable to have the forwarding logic be fairly uniform across different protocol versions.  Right now, we have two almost completely different protocols.

ekinnear:
Capsule design team output. Minimal changes to HTTP/3, which was already using capsules and is now joined by [HTTP/2](https://github.com/ietf-wg-webtrans/draft-webtransport-http2/pull/58).

Same content as [draft PR](), thank you to those who have already reviewed!

Closes #22
Closes #34 
Closes #66

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -452,7 +452,7 @@ Reference:
 
 ## HTTP/3 SETTINGS Parameter Registration
 
-The following entry is added to the "HTTP/3 Settings" registry established by
+The following entries are added to the "HTTP/3 Settings" registry established by
 [HTTP3]:
 
 The `SETTINGS_ENABLE_WEBTRANSPORT` parameter indicates that the specified
@@ -465,6 +465,26 @@ Setting Name:
 Value:
 
 : 0x2b603742
+
+Default:
+
+: 0
+
+Specification:
+
+: This document
+
+The `SETTINGS_WEBTRANSPORT_MAX_SESSIONS` parameter indicates that the specified
+HTTP/3 server is WebTransport-capable and the number of concurrent sessions
+it is willing to receive.
+
+Setting Name:
+
+: WEBTRANSPORT_MAX_SESSIONS
+
+Value:
+
+: 0x2b603743
 
 Default:
 
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -472,11 +472,11 @@ before sending or processing any WebTransport traffic.  When multiple versions
 are supported by both of the peers, the most recent version supported by both
 is selected.
 
-The data exchanged over the CONNECT stream is transmitted across
-intermediaries, and thus cannot be versioned using a SETTINGS parameter.  To
-indicate support for different versions of the protocol defined in this draft,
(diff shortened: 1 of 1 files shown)


vasilvv:
Does this actually close #22 with a no-op, or are we supposed to discuss that at 115?

Otherwise LGTM.

ekinnear:
It closes #22 as a no-op with the exception of one bullet point that we did want to address, that got split out into #85, which we'll discuss at 115 separately from this PR.

DavidSchinazi:
[Consensus declared](https://mailarchive.ietf.org/arch/msg/webtransport/shBK8WFsC9aWihBCy--e-7kKhRc/)

PR #91, opened 2022-12-16
Remove SETTINGS_MAX_WEBTRANSPORT_SESSIONS

ekinnear:
Doing this all together later, so keeping this out for now


ekinnear:
(Discussed with chairs and editors, we know there are outstanding changes, but leaving this for now)

PR #94, opened 2023-02-18
Relax the Origin header requirement

### LINKED ISSUE #92: Why is use of the Origin header field a MUST
In draft 03 the following requirement is made

> An Origin header [[RFC6454](https://www.rfc-editor.org/rfc/rfc6454)] MUST be provided within the request.

WebTransport's closest equivalent in this regard is WebSockets, which doesn't mandate the origin header; see https://www.rfc-editor.org/rfc/rfc6455#section-1.3

This requirement makes it weird to use for user agents that have no concept of an origin model. So the MUST level seems a bit overzealous.

vasilvv:
This is vaguely based on the similar text in the WebSocket RFC.

Fixes #92.


LPardue (approved):
Lgtm, totally makes sense to keep this for browsers.

LPardue (approved):

DavidSchinazi (approved):

aboba (approved):

ekinnear (approved):

PR #95, opened 2023-02-18
Document how to close both the session and the underlying connection

### LINKED ISSUE #89: race condition when closing the underlying QUIC connection
A common use (at least for non-pooled WebTransport connections) will be to close the underlying QUIC connection when closing the WebTransport session.

This is inherently racy: The sender will send both a CLOSE_WEBTRANSPORT_SESSION capsule and a QUIC CONNECTION_CLOSE frame (unless you wait for the peer's FIN on the request stream, but why would you?). It's undetermined which one of the two will be received / processed first by the receiver.

This is not a problem if the sender uses the same error code and error message for the CONNECTION_CLOSE frame and for the CLOSE_WEBTRANSPORT_SESSION capsule. Assuming the QUIC stack properly exposes these values on its API, the receiver will be able to learn error code and error message.

Should we add that as a recommendation, or even a requirement?

vasilvv:
This is mostly based on what Chrome currently does in production.

Fixes #89.

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -420,7 +420,7 @@ CLOSE_WEBTRANSPORT_SESSION capsule that has an error code of 0 and an empty
 error string.
 
 In some scenarios, an endpoint might want to send a CLOSE_WEBTRANSPORT_SESSION
-with detailed close information, and then immediately close the underlying QUIC
+with detailed close information and then immediately close the underlying QUIC
 connection.  If the endpoint were to do both of those simultaneously, the peer
 could potentially receive the CONNECTION_CLOSE before receiving the
 CLOSE_WEBTRANSPORT_SESSION, thus never receiving the application error data
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -426,7 +426,7 @@ could potentially receive the CONNECTION_CLOSE before receiving the
 CLOSE_WEBTRANSPORT_SESSION, thus never receiving the application error data
 contained in the latter.  To avoid this, the endpoint SHOULD wait until either
 all of the data on the CONNECT stream is acknowledged, or three times the PTO
-interval ({{!RFC9002}}, Section 6.2) elapses, before sending the
+interval ({{!RFC9002, Section 6.2}}) elapses, before sending the
 CONNECTION_CLOSE.
 
 # Negotiating the Draft Version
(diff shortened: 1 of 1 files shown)


martinthomson (commented):
My first observation here is that we probably want to treat a closing signal akin to reliable, ordered application data in this context.  That is, we want to provide the same delivery guarantees for the close on the stream/session as the last message that is sent on a stream.

To that end, we might seek to retain connections until data is acknowledged.  That's a general requirement that is not specific to the close signal.  The only reason that close is special is twofold:

1. It is highly desirable that close be infallible.  That is, the operation cannot fail.
2. Closing a session is often followed by closing the connection.

Though on the latter point, maybe it is better if a site is not directly aware of the closure of the underlying connection, any more than it is aware of the operation of a garbage collector.

So I don't think that this is the right framing for this change.  

I'm split on what the right outcome might be.  If the goal is to execute a close that is infallible, then we probably shouldn't add any sort of reliability.  In that case, an application can execute its own shutdown, gaining any confirmation it needs with its own messages.  Close messages at this layer become best-effort.

(This is orthogonal to the use of the close message for connection- and session-management purposes.  We've already decided that the session close message is useful, particularly in the case where you have multiple sessions sharing a connection.)

Alternatively, we provide a shutdown capability that builds on the closing messages.  This can fail.  For this, I'm having trouble with is that acknowledgment state is not something that we guarantee at the application layer.  Maybe [this](https://quicwg.org/base-drafts/rfc9000.html#section-2.4-6) might be used to monitor for a state transition to the [Data Recvd](https://quicwg.org/base-drafts/rfc9000.html#fig-stream-send-states) state, but I'm not sure if this is something that can be reliably implemented.

aboba (approved):

draft-ietf-webtrans-http3.md
```diff
@@ -419,6 +419,16 @@ capsule SHALL be semantically equivalent to terminating it with a
 CLOSE_WEBTRANSPORT_SESSION capsule that has an error code of 0 and an empty
 error string.
 
+In some scenarios, an endpoint might want to send a CLOSE_WEBTRANSPORT_SESSION
+with detailed close information, and then immediately close the underlying QUIC
```
ekinnear, line 423:
```suggestion
with detailed close information and then immediately close the underlying QUIC
```

DavidSchinazi:
Chair: discussed during editor's meeting, plan is to remove text about 3 PTO and note that delivering other open streams is best-effort. Also add rationale and mention that this aligns with what QUIC does - this message has similar delivery guarantees as the close_connection in QUIC

martinthomson (approved):

Issue #81, opened 2022-08-04
Indefinite-length frames are too underspecified


LPardue:
I'm curious how anyone has got handling of the WEBTRANSPORT_STREAM frame working based on strict following of the specs. 

WT over H3 says:

> A client can create a bidirectional stream using a special indefinite-length HTTP/3 frame that transfers ownership of the stream to WebTransport.

H3 says in https://www.rfc-editor.org/rfc/rfc9114.html#name-frame-layout:

> All frames have the following format
> HTTP/3 Frame Format {
  Type (i),
  Length (i),
  Frame Payload (..),
}

snip

> Each frame's payload MUST contain exactly the fields identified in its description. A frame payload that contains additional bytes after the identified fields or a frame payload that terminates before the end of the identified fields MUST be treated as a [connection error](https://www.rfc-editor.org/rfc/rfc9114.html#errors) of type [H3_FRAME_ERROR](https://www.rfc-editor.org/rfc/rfc9114.html#H3_FRAME_ERROR). In particular, redundant length encodings MUST be verified to be self-consistent; see [Section 10.8](https://www.rfc-editor.org/rfc/rfc9114.html#frame-parsing).

WT over h3 says of the `WEBTRANSPORT_STREAM` frame:

> The format of the frame SHALL be the frame type, followed by the session ID, encoded as a variable-length integer, followed by the user-specified stream data ([Figure 2](https://ietf-wg-webtrans.github.io/draft-ietf-webtrans-http3/draft-ietf-webtrans-http3.html#fig-bidi-client)). The frame SHALL last until the end of the stream.

So it appears that WT is trying to omit the length field to imply an Indefinate length, but H3 doesn't allow that. I'd expect a standard H3 frame parser to read the first N bytes of the session ID and incorrectly parse that as the length. I don;t think you can special case this frame, because your breaking the H3 contract that unknown frame types have a correct length that allows them to be read and discarded.

One option (lets call it A) would be to say that `WEBTRANSPORT_STREAM` has a length field of MAX_VAL and then create a new caveat that if you're a WebTransport supporting endpoint, you don't need to create a connection error the stream if it is FIN'd before that length is reached. However, that has the unfortunate problem of consuming 8 bytes at the start of every stream.

Another option (B) might be to say that the `WEBTRANSPORT_STREAM` frame is fixed length as so:

> WEBTRANSPORT_STREAM Frame {
  Type (i) = 0x41,
  Length (i),
[...]

PR #96, opened 2023-02-19
Clarify the nature of WEBTRANSPORT_STREAM

### LINKED ISSUE #48: Unidirectional stream framing
WT uni streams are currently defined as

`Stream Type | Preface (Session ID) | Unframed`

This was awkward to implement, and asymmetric with how client-initiated bidi WT streams are implemented (and now server-initiated bidi -- with a frame).  

I'd like to propose we change WT uni streams to 

`Stream Type | TV Frame` 

or perhaps 

`Stream Type | Preface | TV Frame`

### LINKED ISSUE #71: Bidirectional Stream Frame Ordering Requirements
The draft says:
> WebTransport endpoints can initiate bidirectional streams by opening an HTTP/3 bidirectional stream and sending an HTTP/3 frame with type WEBTRANSPORT_STREAM (type=0x41).

It is unclear if it is allowed to send other frames before sending the WEBTRANSPORT_STREAM frame. I can't think of any currently defined HTTP/3 frame that would make sense to send here, but that might not be the case with future extensions. At the very least, it would be nice of grease frames could be sent.

vasilvv:
Also ban sending it anywhere that's not the beginning of the stream.

Fixes #71
Potentially also fixes #48 and #81

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -243,11 +243,10 @@ by the user-specified stream data ({{fig-unidi}}).
 ## Bidirectional Streams
 
 WebTransport endpoints can initiate bidirectional streams by opening an HTTP/3
-bidirectional stream and sending an HTTP/3 frame with type
-`WEBTRANSPORT_STREAM` (type=0x41).  The format of the frame SHALL be the frame
-type, followed by the session ID, encoded as a variable-length integer,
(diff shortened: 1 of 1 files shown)


vasilvv:
cc @LPardue

DavidSchinazi (approved):

ekinnear (approved):

PR #103, opened 2023-03-08
Expand on HTTP/3 extension aspects

LPardue:
By switching from Frames to special signal values at byte position 0,
we are really extending HTTP/3 rules. So this is an attempt to paint
that out more cleanly and tidy up some loose ends or stale text in the
draft.

The outcome is that we make it clear that WebTransport streams never use
frames. Instead, they are always a sequence of some preamble at the
start of a stream, followed by a session ID, followed by user-specified
stream data. This makes it more obvious that endpoints should never send
0x41 such that the peer could interpret it as being a frame on any
frame-bearing stream (like the control stream). If you understand
WebTransport but didn't enable it, then receiving 0x41 is also bad, so
blow up.

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -83,7 +83,8 @@ path value (defined in {{!RFC3986}} Sections 3.2 and 3.3 correspondingly).
 
 When an HTTP/3 connection is established, both the client and server have to
 send a SETTINGS_ENABLE_WEBTRANSPORT setting in order to indicate that they both
-support WebTransport over HTTP/3.
+support WebTransport over HTTP/3.  This process also negotiates the use of
+additional HTTP/3 extensions.
 
 WebTransport sessions are initiated inside a given HTTP/3 connection by the
 client, who sends an extended CONNECT request {{!RFC8441}}.  If the server
@@ -96,10 +97,10 @@ be further referred to as a *Session ID*.
 After the session is established, the peers can exchange data using the
 following mechanisms:
 
-* A client can create a bidirectional stream using a special indefinite-length
-  HTTP/3 frame that transfers ownership of the stream to WebTransport.
-* A server can create a bidirectional stream, which is possible since HTTP/3
-  does not define any semantics for server-initiated bidirectional streams.
+* A client can create a bidirectional stream and transfer its ownership to
+  WebTransport by providing a special signal in the first bytes.
+* A server can create a bidirectional stream and transfer its ownership to
+  WebTransport by providing a special signal in the first bytes..
 * Both client and server can create a unidirectional stream using a special
   stream type.
 * A datagram can be sent using HTTP Datagrams {{!HTTP-DATAGRAM=RFC9297}}.
@@ -109,7 +110,7 @@ closed.
 
 # Session Establishment
 
-## Establishing a Transport-Capable HTTP/3 Connection
+## Establishing a Transport-Capable HTTP/3 Connection {#establishing}
 
 In order to indicate support for WebTransport, both the client and the server
 MUST send a SETTINGS_ENABLE_WEBTRANSPORT value set to "1" in their SETTINGS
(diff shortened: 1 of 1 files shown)


vasilvv (approved):
LGTM modulo the following:
- I don't think the signal values make sense in a separate draft since their use is inherently tied to WebTransport.
- Some minor wording things.

I will merge this as-is and fix the editorial things in a follow-up commit.

vasilvv:
Ah, noticed I can't actually merge this because conflicts.  I'll leave the comments instead.

draft-ietf-webtrans-http3.md
```diff
@@ -84,7 +84,8 @@ path value (defined in {{!RFC3986}} Sections 3.2 and 3.3 correspondingly).
 
 When an HTTP/3 connection is established, both the client and server have to
 send a SETTINGS_ENABLE_WEBTRANSPORT setting in order to indicate that they
-both support WebTransport over HTTP/3.
+both support WebTransport over HTTP/3. This process also negotiates the use of
+additional HTTP/3 extensions.
```
vasilvv, line 88: "both support WebTransport over HTTP/3, including the necessary extensions to the HTTP/3 syntax".

draft-ietf-webtrans-http3.md
```diff
@@ -231,56 +232,55 @@ type, followed by the session ID, encoded as a variable-length integer, followed
...
-bidirectional stream and then immediately sending a special signal value 0x41,
-followed by the associated session ID, both encoded as a variable-length
-integer; the rest of the stream is the application payload of the WebTransport
-stream ({{fig-bidi-client}}).
+All client-initiated bidirectional streams are reserved by HTTP/3 as request
+streams, which are a sequence of HTTP/3 frames with a variety of rules; see
+{{Sections 4.1 and 6.1 of HTTP3}}.
+
+WebTransport extends HTTP/3 to allow clients to declare and use alternative
+request stream rules.  Once WebTransport has been established
```
vasilvv, line 248: I would prefer to use "syntax" instead of "rules" here and elsewhere.

draft-ietf-webtrans-http3.md
```diff
@@ -231,56 +232,55 @@ type, followed by the session ID, encoded as a variable-length integer, followed
...
+({{establishing}}), a client can send a special signal value, encoded as a
```
vasilvv, line 249: Note that the client may actually send the streams before establishing a transport -- what's required here is that the sender knows that the peer supports WT.

DavidSchinazi:
@vasilvv can you fix the conflict and then merge?

draft-ietf-webtrans-http3.md
```diff
@@ -231,56 +232,55 @@ type, followed by the session ID, encoded as a variable-length integer, followed
...
+how the remaining bytes on the stream are used.
+
+WebTransport extends HTTP/3 by defining rules for all server-initiated
+bidirectional streams.  Once WebTransport has been established
+({{establishing}}), a server can open a bidirectional stream and SHALL send a
+special signal value, encoded as a variable-length integer, as the first bytes
+of the stream in order to indicate how the remaining bytes on the stream are
+used.
+
+TODO: consider extracting special signal values into a separate draft;
```
ekinnear, line 260: Kind of sad to add a TODO here vs. just having an issue for it?

ekinnear (approved):

draft-ietf-webtrans-http3.md
ekinnear: (Just if already fixing conflicts)

vasilvv, line 260: I'm not sure the special values make sense in a separate draft, since they are WT specific.  The old TODO was referring to a separate draft for defining syntax of server-initiated bidi streams.

LPardue: This is effectively just porting the old TODO that existed into this. I agree it could actually be an issue, or just removed if we don't want to do it (as Victor suggested, making a standalone HTTP extension with no context seems a bit problematic)

vasilvv:
> @vasilvv can you fix the conflict and then merge?

Since this is mostly editorial, I don't see the need to rush it.

draft-ietf-webtrans-http3.md
ekinnear: Fair enough 😃

draft-ietf-webtrans-http3.md
LPardue: RFC9114 doesn't declare stream types as syntax though. And the problem is you can send things with the correct byte layout but in the wrong order, which violates the rules set out in 9114 Section 4.1. Specifically, I used `rule(s)` as a direct reference to https://www.rfc-editor.org/rfc/rfc9114.html#section-4.1-7

> Receipt of an invalid sequence of frames MUST be treated as a [connection error](https://www.rfc-editor.org/rfc/rfc9114.html#errors) of type [H3_FRAME_UNEXPECTED](https://www.rfc-editor.org/rfc/rfc9114.html#H3_FRAME_UNEXPECTED). In particular, a [DATA](https://www.rfc-editor.org/rfc/rfc9114.html#frame-data) frame before any [HEADERS](https://www.rfc-editor.org/rfc/rfc9114.html#frame-headers) frame, or a [HEADERS](https://www.rfc-editor.org/rfc/rfc9114.html#frame-headers) or [DATA](https://www.rfc-editor.org/rfc/rfc9114.html#frame-data) frame after the trailing [HEADERS](https://www.rfc-editor.org/rfc/rfc9114.html#frame-headers) frame, is considered invalid. Other frame types, especially unknown frame types, might be permitted subject to their own rules; see [Section 9](https://www.rfc-editor.org/rfc/rfc9114.html#extensions).

draft-ietf-webtrans-http3.md
LPardue: good point!

draft-ietf-webtrans-http3.md
LPardue: Since you've said the TODO doesn't carry over I'll remove it. 

I tend to agree that trying to make this a generic HTTP/3 extension in the abstract is probably not going to help people but that is something the HTTP WG should be consulted on too. A TODO isn't a good way to track that process though.

draft-ietf-webtrans-http3.md
```diff
@@ -240,56 +241,53 @@ type, followed by the session ID, encoded as a variable-length integer, followed
...
-  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-  |                        Session ID (i)                       ...
-  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-  |                         Stream Body                         ...
-  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+Signal Value (i) = 0x41,
+Session ID (i),
+Stream Body (..),
 ~~~~~~~~~~
 {: #fig-bidi-client title="Bidirectional WebTransport stream header"}
```
huitema, line 280: Shouldn't there be some curly braces, as in:
~~~
Bidirectional Stream {
    Signal Value (i) = 0x41,
    Session ID (i),
    Stream Body (..)
}
~~~

vasilvv: Indeed (both here and in the diagram above).

DavidSchinazi:
Chair: discussed during editor's meeting, we'll merge this as-is and keep #107 open to track the diagrams.

PR #105, opened 2023-03-11
Rewrap

ekinnear:
Noticed we had a few oddly long lines, so I hit the rewrap button.


DavidSchinazi (approved):
What tool do you use that has a rewrap button?

vasilvv (approved):

ekinnear:
> What tool do you use that has a rewrap button?

This was Sublime, but many others will attempt to do the same :)

PR #106, opened 2023-03-11
Require SETTINGS_H3_DATAGRAM and max_datagram_frame_size TP

### LINKED ISSUE #84: WT over HTTP/3 should require capsule support and datagram support explicitly
Per discussion at IETF 114, we should explicitly require support for each layer of capabilities that are needed. 

There was also a discussion about whether or not we needed to require datagram support, since requiring capsules technically means that you could use a datagram capsule and implementations may exist that do not want to have datagram support at all.

Documenting some of that discussion in replies to this issue, as well.

ekinnear:
Require SETTINGS_H3_DATAGRAM and max_datagram_frame_size TP

Closes #84

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -132,6 +132,16 @@ concurrent sessions it is willing to receive.  The default value for the
 SETTINGS_MAX_WEBTRANSPORT_SESSIONS parameter is "0", meaning that the server is
 not willing to receive any WebTransport sessions.
 
+Because WebTransport over HTTP/3 requires support for HTTP/3 datagrams and the
+Capsule Protocol, both the client and the server MUST indicate support for
+HTTP/3 datagrams by sending a SETTINGS_H3_DATAGRAM value set to 1 in their
+SETTINGS frame (see {{Section 2.1.1 of HTTP-DATAGRAM}}).
+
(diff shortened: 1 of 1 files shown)


martinthomson (approved):
Would this be easier to process as a list rather than multiple separate requirements?

DavidSchinazi (approved):
The list text in that comment isn't correct - the server shouldn't close the connection because it's legal for the client to not support datagrams if it only sends GETs and POSTs.

However, the text in this PR is correct. I propose we land the PR as-is and can make editorial changes later.

vasilvv (approved):

ekinnear:
> The server shouldn't close the connection because it's legal for the client to not support datagrams if it only sends GETs and POSTs.

While technically true, I thought we'd said that we wanted to require all the features all of the time, but agreed, let's take that to an editorial issue if needed.

DavidSchinazi:
My point is that the server could support both regular HTTP and WebTransport.
* If a client comes in without webtransport settings and only sends GETs, that's fine.
* If a client comes in with webtransport settings and makes a webtransport request, that's fine.
* If a client comes in without webtransport settings and makes a webtransport request, that's not fine. In that case the server needs to reject the webtransport request, not kill the connection.

ekinnear:
Ah, that makes more sense, thank you :)

PR #115, opened 2023-04-17
Extend the error space for reset codes to 32 bits

### LINKED ISSUE #90: increase WebTransport error code space
The current draft specifies that there are 256 error codes that applications running on top of WebTransport can use. That's a very small number, and I'm currently struggling with mapping some application protocols onto WebTransport, because this space is way too small.

I'd like to explore if we can increase this space. 16 bit would be the absolute minimum, 32 bit would be nice, but maybe we can even achieve the full 62 bit that QUIC gives us.

The draft says:
> Since WebTransport shares the error code space with HTTP/3, WebTransport application errors for streams are limited to an unsigned 8-bit integer, assuming values between 0x00 and 0xff.

There's a few problem with that statement:
* It's questionable if WebTransport really shares the error code space with HTTP/3. Arguably, WebTransport takes over the stream from HTTP/3 after the header of the WEBTRANSPORT_STREAM (for bidirectional streams), and after the unidirectional stream type (for unidirectional streams), although technically the respective frame stretches to the end of the stream.
     * This is even more true once we resolve #77, such that stream resets wouldn't end up in limbo when reset before the session ID was read by the receiver.
     * This would allow us to use the entire 62 bit space.
* Even if WebTransport did share the error code space with HTTP/3, it's not clear why this would limit it to an 8-bit integer. HTTP/3 itself uses the the full 62 bit space allowed by QUIC, so it doesn't seem excessive to reserve more than 8 bits out of that space for WebTransport. Even if we consumed 32 bits out of that space, this would barely make a dent in the number of unclaimed HTTP/3 error codes.

One thing to keep in mind is that when mapping WebTransport onto HTTP/2, we only have access to 32 bit error code space. It might be desirable to use the same error code space both mappings of WebTransport.

vasilvv:
Fixes #90.

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -299,20 +299,20 @@ implementation to the application.
 
 A WebTransport application SHALL provide an error code for those operations.
 Since WebTransport shares the error code space with HTTP/3, WebTransport
-application errors for streams are limited to an unsigned 8-bit integer,
-assuming values between 0x00 and 0xff.  WebTransport implementations SHALL
-remap those error codes into an error range where 0x00 corresponds to
-0x52e4a40fa8db, and 0xff corresponds to 0x52e4a40fa9e2.  Note that there are
-code points inside that range of form "0x1f * N + 0x21" that are reserved by
-{{Section 8.1 of HTTP3}}; those have to be accounted for when mapping the error
-codes by skipping them (i.e. the two HTTP/3 error codepoints adjacent to a
-GREASE codepoint would map to two adjacent WebTransport application error
-codepoints).  An example pseudocode can be seen in
-{{fig-remap-errors}}.
+application errors for streams are limited to an unsigned 32-bit integer,
+assuming values between 0x00000000 and 0xffffffff.  WebTransport
+implementations SHALL remap those error codes into an error range where
+0x00000000 corresponds to 0x52e4a40fa8db, and 0xffffffff corresponds to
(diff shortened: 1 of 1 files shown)


draft-ietf-webtrans-http3.md
```diff
@@ -299,20 +299,20 @@ implementation to the application.
...
-0x52e4a40fa8db, and 0xff corresponds to 0x52e4a40fa9e2.  Note that there are
-code points inside that range of form "0x1f * N + 0x21" that are reserved by
-{{Section 8.1 of HTTP3}}; those have to be accounted for when mapping the error
-codes by skipping them (i.e. the two HTTP/3 error codepoints adjacent to a
-GREASE codepoint would map to two adjacent WebTransport application error
-codepoints).  An example pseudocode can be seen in
-{{fig-remap-errors}}.
+application errors for streams are limited to an unsigned 32-bit integer,
+assuming values between 0x00000000 and 0xffffffff.  WebTransport
+implementations SHALL remap those error codes into an error range where
```
marten-seemann, line 304: This is not really normative. Maybe better: error codes are remapped …

DavidSchinazi (approved):
Text LGTM (though I didn't review the math to make sure the values line up)

marten-seemann (approved):
I made the change in webtransport-go (https://github.com/quic-go/webtransport-go/pull/74) and it looks like the math is correct.

ekinnear (approved):

afrind (approved):

PR #118, opened 2023-04-18
Update stream format diagrams

### LINKED ISSUE #107: Please use RFC 9000 style message formats instead of misleading octet graphs
I just got mislead by the way the webtransport draft is written. It includes drawings like:

~~~
     0                   1                   2                   3
     0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
     |                           0x54 (i)                          ...
     +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
~~~

Looking at the graphic, you get the impression that the first byte of the unidir stream is 0x54. Of course, that's incorrect, the varint is encoded on two bytes, 0x4054. But the confusion would not happen if the text used RFC 9000 style notations, as in:

~~~
unidir-stream {
    signal-value(i),
    session-id(i),
    stream-body(...)
}
~~~

It would be nce to fix that!

By the way, there is a nice interop test. What if the session ID was encoded as 0x8000000054 ?

ekinnear:
Update stream format diagrams

Closes #107

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -241,14 +241,16 @@ with an H3_ID_ERROR error code.
 ## Unidirectional streams
 
 WebTransport endpoints can initiate unidirectional streams.  The HTTP/3
-unidirectional stream type SHALL be 0x54.  The body of the stream SHALL be the stream
-type, followed by the session ID, encoded as a variable-length integer, followed
-by the user-specified stream data ({{fig-unidi}}).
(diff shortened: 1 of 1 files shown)


LPardue:
thanks!

Issue #102, opened 2023-03-08
Priorities and Pooling


afrind:
I made this comment on a PR related to the Warp draft for moq: https://github.com/kixelated/warp-draft/pull/98

How are streams across different WebTransport sessions prioritized? The WebTransport CONNECT request can have an HTTP Priority header -- one logical interpretation is that all streams belonging to that session be prioritized "under" that top level prioritization. eg: if I have two sessions at the same urgency and incremental on, session A has 3 streams and session B has 1, then I might expect the scheduler to round-robin at the session level (rather than the QUIC stream level), splitting the bandwidth 50/50 between A and B. Is that correct?  For WebTransport over H2, that is the most natural (or only) interpretation.

DavidSchinazi:
Let's discuss at 116. This could be a new feature in the drafts or it could be tackled by a future extension.

LPardue:
FWIW, RFC 9218 sort of recommends sharing already for conventional CONNECT; https://www.rfc-editor.org/rfc/rfc9218.html#name-scheduling-and-the-connect-

I keep hearing that people want strict ordering guarantees among streams. Are you saying that you'd want a fair share of bandwidth but strict ordering inside that fair share?

afrind:
> fair share of bandwidth but strict ordering inside that fair share

I think I want to apply the HTTP prioritization per WT session (so one session could entirely block another, if they have different urgency, eg) but additional ability to prioritize streams within that session, which could be strict ordering, incremental or some combination.

afrind:
Discussed the following at IETF 116

Option 1: Nothing, you get what you get and don’t throw a fit
Option 2: Streams are prioritized globally, be careful what you pool
Option 3: WebTransport sessions are round robin, any stream prioritization is scoped within a session 
	What about HTTP request streams?
Option 4: WebTransport sessions and HTTP request streams are prioritized by Priority header, any WebTransport stream prioritization is scoped within a session

The sense I got from the room was that folks preferred Option 4.

PR #121, opened 2023-05-02
Explain how Priority interacts with WebTransport sessions

afrind:
Fixes: #102

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -224,28 +224,30 @@ session have alternative mechanisms:
 
 ## Prioritization
 
-WebTransport does not provide any priority signalling mechanism for streams and
-datagrams within a WebTransport session; such mechanisms can be defined by
-application protocols using WebTransport.
+WebTransport sessions are initiated using extended CONNECT. While {{Section 11
+of !RFC9218}} describes how extensible priorities can be applied to data sent on
+a CONNECT stream, WebTransport extends the types of data that are exchanged in
+relation to the request and response, which requires additional considerations.
 
-The Priority header {{!RFC9218}} MAY be set in a WebTransport CONNECT request or
-response; clients MAY also reprioritize WebTransport sessions by sending
-PRIORITY_UPDATE frames ({{Section 7 of RFC9218}}).  These priority signals
-SHOULD be included by the recipient as input to the scheduling process for all
-data it sends in the enclosing WebTransport session, including Capsules,
-WebTransport streams and datagrams.
+WebTransport CONNECT requests and responses MAY contain the Priority header
+field ({{Section 5 of RFC9218}}); clients MAY reprioritize by sending
+PRIORITY_UPDATE frames ({{Section 7 of RFC9218}}). In extension to {{RFC9218}},
+it is RECOMMENDED that clients and servers apply the scheduling guidance in both
+{{Section 9 of RFC9218}} and {{Section 10 of RFC9218}} for all data that they
+send in the enclosing WebTransport session, including Capsules, WebTransport
+streams and datagrams. WebTransport does not provide any priority signaling
+mechanism for streams and datagrams within a WebTransport session; such
+mechanisms can be defined by application protocols using WebTransport.  It is
+RECOMMENDED that such mechanisms only affect scheduling within a session and not
+scheduling of other data on the same HTTP/3 connection.
 
-For example, when two WebTransport sessions are established on the same HTTP/3
-connection, each with `Priority: u=1, i`, the scheduler can alternate sending
-data from each session.  If an HTTP request is also sent on this connection with
-`Priority: u=0`, the scheduler would send the response data ahead of any data
-from either WebTransport session.
+The client/server priority merging guidance given in {{Section 8 of RFC9218}}
+also applies to WebTransport session. For example, a client that receives a
+response Priority header field could alter its view of a WebTransport session
+priority and alter the scheduling of outgoing data as a result.
 
-Endpoints MAY set the incremental flag set to false for WebTransport sessions,
-but caution is advised as one session could starve other requests and sessions
-with equal priority.  See {{Section 10 of RFC9218}} for general guidance to
-avoid starvation.  {{Section 11 of RFC9218}} gives guidance the CONNECT method,
-which is also applicable to WebTransport.
+Endpoints that prioritize WebTransport sessions need to consider how they
+interact with other sessions or requests on the same HTTP/3 connection.
 
 # WebTransport Features
 
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -225,23 +225,27 @@ session have alternative mechanisms:
 ## Prioritization
 
 WebTransport does not provide any priority signalling mechanism for streams and
-datagrams within a WebTransport session; such mechanisms will be defined by
+datagrams within a WebTransport session; such mechanisms can be defined by
 application protocols using WebTransport.
 
-{{!RFC9218}} defines the Priority header which MAY be set in a WebTransport
-CONNECT request or response.  This header SHOULD govern the scheduling of all
-data within the WebTransport session, including WebTransport streams and
-datagrams.
+The Priority header {{!RFC9218}} MAY be set in a WebTransport CONNECT request or
+response; clients MAY also reprioritize WebTransport sessions by sending
+PRIORITY_UPDATE frames ({{Section 7 of RFC9218}}).  These priority signals
+SHOULD be included by the recipient as input to the scheduling process for all
+data it sends in the enclosing WebTransport session, including Capsules,
+WebTransport streams and datagrams.
 
 For example, when two WebTransport sessions are established on the same HTTP/3
-connection, each with `Priority: u=1, i`, the scheduler alternates sending data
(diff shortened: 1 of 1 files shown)


draft-ietf-webtrans-http3.md
```diff
@@ -218,6 +218,27 @@ session have alternative mechanisms:
   limiting {{!RFC6585}}.  Unlike the previous method, this signal is directly
   propagated to the application.
 
+## Prioritization
+
+WebTransport does not provide any priority signalling mechanism for streams and
+datagrams within a WebTransport session; such mechanisms will be defined by
```
DavidSchinazi, line 224:
```suggestion
datagrams within a WebTransport session; such mechanisms can be defined by
```

draft-ietf-webtrans-http3.md
```diff
@@ -218,6 +218,27 @@ session have alternative mechanisms:
...
+application protocols using WebTransport.
+
+{{!RFC9218}} defines the Priority header which MAY be set in a WebTransport
```
DavidSchinazi, line 227:
```suggestion
The Priority header ({{!RFC9218}}) MAY be set in a WebTransport
```

draft-ietf-webtrans-http3.md
```diff
@@ -218,6 +218,27 @@ session have alternative mechanisms:
...
+CONNECT request or response.  This header SHOULD govern the scheduling of all
+data within the WebTransport session, including WebTransport streams and
+datagrams.
+
+For example, when two WebTransport sessions are established on the same HTTP/3
+connection, each with `Priority: u=1, i`, the scheduler alternates sending data
+from each session.  If an HTTP request is also sent on this connection with
+`Priority: u=0`, the scheduler would send the response data ahead of any data
```
DavidSchinazi, line 235: This example seems to prescribe an algorithm that can lead to starvation, was that intentional?

draft-ietf-webtrans-http3.md
```diff
@@ -218,6 +218,27 @@ session have alternative mechanisms:
...
+from either WebTransport session.
+
+Non-incremental priorities for WebTransport sessions are allowed, but caution is
```
DavidSchinazi, line 238: We should define what a "non-incremental priority" is

draft-ietf-webtrans-http3.md
LPardue: +1

draft-ietf-webtrans-http3.md
```diff
@@ -218,6 +218,27 @@ session have alternative mechanisms:
...
+## Prioritization
+
+WebTransport does not provide any priority signalling mechanism for streams and
+datagrams within a WebTransport session; such mechanisms will be defined by
+application protocols using WebTransport.
+
+{{!RFC9218}} defines the Priority header which MAY be set in a WebTransport
+CONNECT request or response.  This header SHOULD govern the scheduling of all
+data within the WebTransport session, including WebTransport streams and
+datagrams.
```
LPardue, line 230: It seems like you're making new rules for RFC 9218 here. You don't mention the frame (which can only go in one direction) and you're too ambiguous about how these signals affect the dataflows. When a client sends the signal, what is the prioritization affecting (local streams or remote streams?) and vice versa. 

I'm also not sure how this is supposed to gel with the W3C WebTransport API that doesn't allow control over these fields.

draft-ietf-webtrans-http3.md
LPardue: This is a concrete example of the scheduling guidance in RFC 9218, at least as far as what we say a server should do. We can probably reference that section, which also describes the starvation issues.

draft-ietf-webtrans-http3.md
LPardue: or even just a "may be defined by other specifications", which would keep the door open to defining an actual WebTransport signal.

draft-ietf-webtrans-http3.md
```diff
@@ -218,6 +218,27 @@ session have alternative mechanisms:
...
+
+For example, when two WebTransport sessions are established on the same HTTP/3
+connection, each with `Priority: u=1, i`, the scheduler alternates sending data
```
martinthomson, line 233:
```suggestion
connection, each with the same urgency and incremental delivery (i.e., 
`Priority: u=1, i`), the scheduler alternates sending data
```

draft-ietf-webtrans-http3.md
ekinnear: If I'm understanding correctly, this is saying that you must prioritize everything associated with a given WebTransport session at the same level. Technically, that's not required since we broke out each stream into a "native" H3 stream, but that seems like more of a bug than a feature, since interleaving priorities between streams of different WebTransport sessions seems like a recipe for pain. Instead, we probably want to say that everything must stay within the session, that session has a fixed priority, and if you want to prioritize within the session, it's up to you to send the right data in the right order (see the W3C API for that).

draft-ietf-webtrans-http3.md
```diff
@@ -218,6 +218,31 @@ session have alternative mechanisms:
...
+datagrams within a WebTransport session; such mechanisms can be defined by
+application protocols using WebTransport.
+
+The Priority header {{!RFC9218}} MAY be set in a WebTransport CONNECT request or
+response; clients MAY also reprioritize WebTransport sessions by sending
+PRIORITY_UPDATE frames ({{Section 7 of RFC9218}}).  These priority signals
+SHOULD be included by the recipient as input to the scheduling process for all
+data it sends in the enclosing WebTransport session, including Capsules,
+WebTransport streams and datagrams.
+
```
LPardue, line 233: It's not just the recipient, consider the case where a client sends signal X and something coordinating with the server says Y, the server will likely pick Y due to https://www.rfc-editor.org/rfc/rfc9218.html#section-8

That's what's not clear here. In HTTP the interaction expectations were set out clearly. WebTransport is a bit different and we should make sure everyone understands that. 

So as a concrete strawman, I'd phrase it something like this (and run into your second para example, which is fine as it is.

```suggestion
WebTransport sessions are initiated using the CONNECT method. While  {{Section
11 of !RFC9218}} describes how extensible priorities can be applied to CONNECT,
WebTransport extends the types of data that are exchanged in relation to the
request and response, which requires additional considerations. WebTransport
CONNECT requests and responses MAY contain the Priority header field  ({{Section
5 of RFC9218}}); clients MAY reprioritize by sending PRIORITY_UPDATE frames
({{Section 7 of RFC9218}}). In extension to {{RFC9218}}, it is RECOMMENDED that
the scheduling guidance in both {{Section 9 of RFC9218}} and {{Section 10 of
RFC9218}} is applied by clients and servers for all data that they send in the
enclosing WebTransport session, including Capsules, WebTransport streams and
datagrams. Note that the merging considerations given in {{Section 8 of
RFC9218}} also apply. For example, a client that receives a response Priority
header field could alter its view of a WebTransport session priority and alter
the scheduling of outgoing data as a result. 

Endpoints that prioritize WebTransport sessions need to consider how they
interact with other sessions or requests on the same HTTP/3 connection.
```

afrind: Thanks, overall, your text is much better. 

> it is RECOMMENDED that the scheduling guidance in both {{Section 9 of RFC9218}} and {{Section 10 of RFC9218}} is applied by clients and servers for all data that they send in the enclosing WebTransport session, including Capsules, WebTransport streams and datagrams

I think the bit I want to clarify is that prioritizing WebTransport streams or datagrams (which can be defined elsewhere) is only intended to prioritize them against other data in the same WebTransport session, and the Priority header is what prioritizes the session itself. 

Admittedly, my text didn't convey this concept clearly either.

LPardue: ah got it. I agree on the concept. I think we can highlight that with a little bit more iteration of the text in this section. 

The key point that does need to be covered, if I understand your intent for whole-session priotization,  is that RFC 9218 Section 9 says client MAY apply signals, and in WebTransport you want to upgrade it to a SHOULD.

afrind: Oh actually the key point was the 'dependent streams and datagrams get prioritized within the session and not across sessions'; asking the client to respect the server's preference on send with a SHOULD was more of a target of opportunity, and I could be convinced to leave it as a MAY.  I can contrive a use case where there are pooled sessions and the server wants to elevate the priority of one, or indicate a pooled POST is lower priority than some WT session, but a) I'm not sure how real they are and b) I don't know if the difference between MAY and SHOULD makes any difference.

LPardue: The concepts are tangential to me. I.e.

1. A client SHOULD employ some form of prioritization when pooling (or else you might do something naive and cause starvation)
2. A client SHOULD apply the scheduling guidance in section 9 when sending session data (a deviation from section 8, which was written to be interpreted very openly)
3. A client can send a signal to indicate it's send priority to the origin (or intermediary)
4. A server might want to consider the client priority signal but it doesn't have to
5. A client might want to consider the server priority but it doesn't have to

martinthomson: I would drop 4 & 5.

LPardue (approved):
LGTM

vasilvv (approved):

DavidSchinazi (approved):

aboba (approved):

PR #123, opened 2023-05-10
Tweak how HTTP/3 error codes are presented for use

### LINKED ISSUE #117: Reserved HTTP/3 error codes for Webtransport have a name, for what purpose?
By my understanding, we are reserving a range of error codes for WebTransport applications to use. At the time of writing, the IANA considerations says to reserve these as `H3_WEBTRANSPORT_APPLICATION_<MIN>..H3_WEBTRANSPORT_APPLICATION_<MAX>` (where <MIN> and <MAX> are a little in flux due to the extension of the space in https://github.com/ietf-wg-webtrans/draft-ietf-webtrans-http3/pull/115.

What isn't clear to me is who the intended audience of those error code names are? Applications on top of WebTransport are probably going to want to use their own semantic names for the errors (e.g. MOQ_NO_ERROR, MOQ_AWFUL_THING) or whatever. They could do that by saying something like 

> when condition X happens, close the WebTransport session with MOQ_AWFUL_THING ((WebTransport error code 1)|. 

Or is the intention that they will use a form like:

> > when condition X happens, close the WebTransport session with MOQ_AWFUL_THING ((H3_WEBTRANSPORT_APPLICATION_000001). 

Whatever the preference, we should probably start thinking about how to explain to people consistent sytyling for using WebTransport. So this is possibly or more general question.

LPardue:
This change is editorial. It is not intended to redefine how the error
codes operate but instead clarify their usage and make them more
consistent with how HTTP/3 and QUIC do things.

So, we drop `H3` prefix from the error codes, because this is all about
WebTransport.

Then we make a single label - WEBTRANSPORT_APPLICATION_ERROR - that
covers the entire range of codes that we care about. This is similar to
how QUIC's CRYPTO_ERROR works. And in the part we talk about mapping,
add a little callout to the label similar to https://datatracker.ietf.org/doc/html/rfc9001#section-4.8

Fixes #117

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -316,7 +316,7 @@ WEBTRANSPORT_APPLICATION_ERROR, where 0x00000000 corresponds to 0x52e4a40fa8db,
 and 0xffffffff corresponds to 0x52e5ac983162.  Note that there are code points
 inside that range of form "0x1f * N + 0x21" that are reserved by {{Section 8.1
 of HTTP3}}; those have to be skipped when mapping the error codes
-skipping them (i.e. the two HTTP/3 error codepoints adjacent to a reserved
+(i.e. the two HTTP/3 error codepoints adjacent to a reserved
 codepoint would map to two adjacent WebTransport application error codepoints).
 An example pseudocode can be seen in {{fig-remap-errors}}.
 
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -315,7 +315,7 @@ implementations SHALL remap those error codes into the error range reserved for
 WEBTRANSPORT_APPLICATION_ERROR, where 0x00000000 corresponds to 0x52e4a40fa8db,
 and 0xffffffff corresponds to 0x52e5ac983162.  Note that there are code points
 inside that range of form "0x1f * N + 0x21" that are reserved by {{Section 8.1
-of HTTP3}}; those have to be accounted for when mapping the error codes by
+of HTTP3}}; those have to be skipped when mapping the error codes
 skipping them (i.e. the two HTTP/3 error codepoints adjacent to a reserved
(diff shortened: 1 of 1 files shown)


DavidSchinazi (approved):

vasilvv (approved):

draft-ietf-webtrans-http3.md
```diff
@@ -311,14 +311,14 @@ A WebTransport application SHALL provide an error code for those operations.
...
-"0x1f * N + 0x21" that are reserved by {{Section 8.1 of HTTP3}}; those have to
-be accounted for when mapping the error codes by skipping them (i.e. the two
-HTTP/3 error codepoints adjacent to a reserved codepoint would map to two
-adjacent WebTransport application error codepoints).  An example pseudocode can
-be seen in {{fig-remap-errors}}.
+implementations SHALL remap those error codes into the error range reserved for
+WEBTRANSPORT_APPLICATION_ERROR, where 0x00000000 corresponds to 0x52e4a40fa8db,
+and 0xffffffff corresponds to 0x52e5ac983162.  Note that there are code points
+inside that range of form "0x1f * N + 0x21" that are reserved by {{Section 8.1
+of HTTP3}}; those have to be accounted for when mapping the error codes by
```
ekinnear, line 318:
```suggestion
of HTTP3}}; those have to be skipped when mapping the error codes
```

draft-ietf-webtrans-http3.md
```diff
@@ -311,14 +311,14 @@ A WebTransport application SHALL provide an error code for those operations.
...
+skipping them (i.e. the two HTTP/3 error codepoints adjacent to a reserved
```
ekinnear, line 319:
```suggestion
(i.e. the two HTTP/3 error codepoints adjacent to a reserved
```

ekinnear (approved):

ekinnear:
(Landing suggestions during editor's meeting, as the text was present before/after the changes here)

Issue #125, opened 2023-05-23
Unify on a single setting to indicate support for WebTransport?


ekinnear:
This came up in an editor's call when keeping settings in sync. Several folks thought this was already unified, so filing an issue to cover cleaning this up.

Today, HTTP/2 uses `SETTINGS_WEBTRANSPORT_MAX_SESSIONS` instead of `SETTINGS_ENABLE_WEBTRANSPORT` to indicate support for WebTransport and requires both the client and the server to send the setting with a non-zero value.

HTTP/3 still uses _both_ `SETTINGS_WEBTRANSPORT_MAX_SESSIONS` and `SETTINGS_ENABLE_WEBTRANSPORT`, and at the time we said it was in part due to the fact that "h3 requires `SETTINGS_ENABLE_WEBTRANSPORT` because it changes the format of server-initiated bidi streams" ([context](https://github.com/ietf-wg-webtrans/draft-ietf-webtrans-http2/issues/73#issuecomment-1464720983)).

However, we've said that `SETTINGS_WEBTRANSPORT_MAX_SESSIONS` > 0 is semantically equivalent to `SETTINGS_ENABLE_WEBTRANSPORT`.

If that's the case, then it should be okay to replace that setting with the one by a different name, with otherwise identical language. It's a little bit odd to send `SETTINGS_WEBTRANSPORT_MAX_SESSIONS` from an endpoint that isn't able to have sessions opened to it, but it seems like completely eliminating a setting reduces complexity and the actual string spelling of the setting name isn't that important. That would unify both HTTP/2 and HTTP/3 on a single way to say "yes, I speak WebTransport and you can speak WebTransport to me".

Concretely, the proposal is to replace uses of
`SETTINGS_ENABLE_WEBTRANSPORT` = 1
with
`SETTINGS_WEBTRANSPORT_MAX_SESSIONS` = 1
with no other change in meaning or semantics.

LPardue:
This seems like a good change. 

> Concretely, the proposal is to replace uses of
> SETTINGS_ENABLE_WEBTRANSPORT = 1
> with
> SETTINGS_WEBTRANSPORT_MAX_SESSIONS = 1
> with no other change in meaning or semantics.

I think practically this substitution is more like moving to "greater than 0", to marry up with how the HTTP/2 spec does it.

PR #128, opened 2023-06-09
Redo the version bump

vasilvv:
In a previous PR, I said that I bumped a SETTINGS parameter number to indicate the new draft version.  Turns out, I accidentally changed the WEBTRANSPORT_MAX_SESSIONS codepoint instead.

This PR actually changes the correct one (ENABLE_WEBTRANSPORT). It also replaces "draft-06" with "draft-07", since draft-06 is not really implementable at this state.

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -536,7 +536,7 @@ Setting Name:
 
 Value:
 
-: 0x2b603742
+: 0x9412f086
 
 Default:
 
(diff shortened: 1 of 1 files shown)


DavidSchinazi:
Can we have this PR also resolve #125 ? Let's avoid changing settings multiple times to reduce implementation churn

vasilvv:
> Can we have this PR also resolve #125 ? Let's avoid changing settings multiple times to reduce implementation churn

I don't think we should block solving the easy problem (which is me screwing up and changing the wrong number), and the hard problem (which is merging two settings into one).  I don't think there's implementation churn from this, since draft-06 is not really implementable (at least in backwards-compatible way) because of this problem.

ekinnear:
I thought the other PR wasn't so much a mistake but rather was intentional, since we were talking about changing the setting. So to that end, I'd be tempted to just do #125 (happy to get a PR out for that shortly) and have this just update the text to point at the correct setting to look at.

ekinnear:
Opened #129, so perhaps this would just keep the updates to the changelog and also note the change in setting name?

ekinnear (approved):

PR #129, opened 2023-06-09
Replace SETTINGS_ENABLE_WEBTRANSPORT with SETTINGS_WEBTRANSPORT_MAX_SESSIONS

### LINKED ISSUE #125: Unify on a single setting to indicate support for WebTransport?

ekinnear:
Replace SETTINGS_ENABLE_WEBTRANSPORT with SETTINGS_WEBTRANSPORT_MAX_SESSIONS

Fixes #125

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -121,10 +121,13 @@ than "0", since clients initiate WebTransport sessions, and the actual value is
 not significant.
 
 The client MUST NOT send a WebTransport request until it has received the
-setting indicating WebTransport support from the server.  Similarly, the server
-MUST NOT process any incoming WebTransport requests until the client settings
-have been received, as the client may be using a version of the WebTransport
-extension that is different from the one used by the server.
+setting indicating WebTransport support from the server.  
+
+\[\[RFC editor: please remove the following paragraph before publication.]]
+
+Similarly, the server MUST NOT process any incoming WebTransport requests until
+the client settings have been received, as the client may be using a version of
+the WebTransport extension that is different from the one used by the server.
 
 Because WebTransport over HTTP/3 requires support for HTTP/3 datagrams and the
 Capsule Protocol, both the client and the server MUST indicate support for
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -116,7 +116,9 @@ In order to indicate support for WebTransport, both the client and the server
 MUST send a SETTINGS_WEBTRANSPORT_MAX_SESSIONS value greater than "0" in their
 SETTINGS frame.  The default value for the SETTINGS_WEBTRANSPORT_MAX_SESSIONS
 parameter is "0", meaning that the endpoint is not willing to receive any
-WebTransport sessions.
+WebTransport sessions.  Note that the client only needs to send a value greater
+than "0", since clients initiate WebTransport sessions, and the actual value is
+not significant.
 
 The client MUST NOT send a WebTransport request until it has received the
 setting indicating WebTransport support from the server.  Similarly, the server
@@ -138,9 +140,11 @@ of !QUIC-DATAGRAM=RFC9221}}).
 
 {{!RFC8441}} defines an extended CONNECT method in Section 4, enabled by the
 SETTINGS_ENABLE_CONNECT_PROTOCOL setting.  That setting is defined for HTTP/3
-by {{!RFC9220}}.  An endpoint supporting WebTransport over HTTP/3 MUST send
-both the SETTINGS_WEBTRANSPORT_MAX_SESSIONS setting with a value greater
-than "0" and the SETTINGS_ENABLE_CONNECT_PROTOCOL setting with a value of "1".
+by {{!RFC9220}}.  A client supporting WebTransport over HTTP/3 MUST send the
+SETTINGS_WEBTRANSPORT_MAX_SESSIONS setting with a value greater than "0". A
+server supporting WebTransport over HTTP/3 MUST send both the
+SETTINGS_WEBTRANSPORT_MAX_SESSIONS setting with a value greater than "0" and
+the SETTINGS_ENABLE_CONNECT_PROTOCOL setting with a value of "1".
 
 ## Creating a New Session
 
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -82,9 +82,9 @@ WebTransport servers in general are identified by a pair of authority value and
 path value (defined in {{!RFC3986}} Sections 3.2 and 3.3 correspondingly).
 
 When an HTTP/3 connection is established, both the client and server have to
-send a SETTINGS_ENABLE_WEBTRANSPORT setting in order to indicate that they both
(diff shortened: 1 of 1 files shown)


draft-ietf-webtrans-http3.md
```diff
@@ -82,9 +82,9 @@ WebTransport servers in general are identified by a pair of authority value and
 path value (defined in {{!RFC3986}} Sections 3.2 and 3.3 correspondingly).
 
 When an HTTP/3 connection is established, both the client and server have to
-send a SETTINGS_ENABLE_WEBTRANSPORT setting in order to indicate that they both
-support WebTransport over HTTP/3.  This process also negotiates the use of
-additional HTTP/3 extensions.
+send a SETTINGS_WEBTRANSPORT_MAX_SESSIONS setting in order to indicate that
+they both support WebTransport over HTTP/3.  This process also negotiates the
```
DavidSchinazi, line 86: Unrelated to this PR, but `This process also negotiates the use of additional HTTP/3 extensions.` is no longer true now that we require negotiating each extension separately. Can we just remove that sentence?

draft-ietf-webtrans-http3.md
```diff
@@ -113,23 +113,17 @@ closed.
...
+MUST send a SETTINGS_WEBTRANSPORT_MAX_SESSIONS value greater than "0" in their
+SETTINGS frame.  The default value for the SETTINGS_WEBTRANSPORT_MAX_SESSIONS
+parameter is "0", meaning that the endpoint is not willing to receive any
+WebTransport sessions.
 
 The client MUST NOT send a WebTransport request until it has received the
 setting indicating WebTransport support from the server.  Similarly, the server
 MUST NOT process any incoming WebTransport requests until the client settings
-have been received, as the client may be using a version of WebTransport
+have been received, as the client may be using a version of the WebTransport
```
DavidSchinazi, line 124: Should this paragraph be in a "to be removed by RFC editor" portion?

draft-ietf-webtrans-http3.md
```diff
@@ -145,8 +139,8 @@ of !QUIC-DATAGRAM=RFC9221}}).
 {{!RFC8441}} defines an extended CONNECT method in Section 4, enabled by the
 SETTINGS_ENABLE_CONNECT_PROTOCOL setting.  That setting is defined for HTTP/3
 by {{!RFC9220}}.  An endpoint supporting WebTransport over HTTP/3 MUST send
-both the SETTINGS_ENABLE_WEBTRANSPORT setting and the
-SETTINGS_ENABLE_CONNECT_PROTOCOL setting with values set to "1".
+both the SETTINGS_WEBTRANSPORT_MAX_SESSIONS setting with a value greater
+than "0" and the SETTINGS_ENABLE_CONNECT_PROTOCOL setting with a value of "1".
```
DavidSchinazi, line 143: `SETTINGS_ENABLE_CONNECT_PROTOCOL` is only sent by servers

DavidSchinazi (approved):
Overall this PR is great, my comments apply to the preexisting text but it'd be nice to clean all this up once and for all

vasilvv:
I don't believe this addresses the main reason we've not done this yet, which is the fact that SETTINGS_WEBTRANSPORT_MAX_SESSIONS is only sent from the server to the client, making it not particularly useful for version negotiation.

draft-ietf-webtrans-http3.md
vasilvv: I believe Lucas added this in the PR related to WEBTRANSPORT_STREAM, so we may want to clarify it instead of removing.

ekinnear:
> SETTINGS_WEBTRANSPORT_MAX_SESSIONS is only sent from the server to the client, making it not particularly useful for version negotiation.

The intent was that this is rectified by the new text in this PR, if that doesn't seem to be the case, then let's fix it!

For example:
```
When an HTTP/3 connection is established, both the client and server have to
send a SETTINGS_WEBTRANSPORT_MAX_SESSIONS setting in order to indicate that
they both support WebTransport over HTTP/3. 
```

DavidSchinazi:
Long term, once this has shipped as RFC and all browsers have dropped support for pre-RFC drafts, is the plan to have clients send SETTINGS_WEBTRANSPORT_MAX_SESSIONS forever or to stop sending it? If the goal is to only do this during the draft period, we should have "to be removed by RFC editor" notes explaining that

draft-ietf-webtrans-http3.md
ekinnear: Makes sense, done in https://github.com/ietf-wg-webtrans/draft-ietf-webtrans-http3/pull/129/commits/6b437df3129ac03d6ef429f5e1dfc8785f78ffdd

draft-ietf-webtrans-http3.md
ekinnear: Good call, updated and added a note about the client's value not mattering much in https://github.com/ietf-wg-webtrans/draft-ietf-webtrans-http3/pull/129/commits/0deb77283cc5599741e7dc74e218a7635deecb3a

ekinnear:
> Long term, once this has shipped as RFC and all browsers have dropped support for pre-RFC drafts, is the plan to have clients send SETTINGS_WEBTRANSPORT_MAX_SESSIONS forever or to stop sending it? If the goal is to only do this during the draft period, we should have "to be removed by RFC editor" notes explaining that

Filed #130 to discuss this

vasilvv (approved):

ekinnear:
Discussed in editor's meeting, moved follow-ups out to an issue and merging.

PR #131, opened 2023-06-27
Clarify that GOAWAY signals drain for all sessions

### LINKED ISSUE #126: Clarify the GOAWAY semantics for WebTransport session
In #76, we clarified that GOAWAY does not stop you from opening new streams or sending more datagrams, but we never indicated whether it has any other semantic effect.  My interpretation (that is at least in part [reflected in the current W3C API](https://w3c.github.io/webtransport/#session-draining)) is that a GOAWAY should be treated as semantically equivalent to DRAIN_WEBTRANSPORT_SESSION on every WebTransport session: it does not actually affect the ability to use those sessions, but the client should attempt to gracefully terminate ASAP.  Does this sound reasonable?

vasilvv:
Fixes #126

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -402,6 +402,9 @@ DRAIN_WEBTRANSPORT_SESSION Capsule {
 }
 ~~~
 
+Receiving an HTTP/3 GOAWAY frame is semantically equivalent to receiving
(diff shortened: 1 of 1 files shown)


ekinnear (approved):

afrind (approved):

martinthomson:
See #132

DavidSchinazi:
#132 seems better than this PR

vasilvv:
Closed in favor of #132.

PR #144, opened 2023-11-07
Add subprotocol negotiation

### LINKED ISSUE #137: Define a header for subprotocol negotiation
To be able to implement w3c/webtransport#536 (see moq-wg/moq-transport#233 for some background, but the problem here is porting applications that rely on ALPN).  Client sends a list of supported subprotocols, server responds with one of the selected.

Possible syntax ideas for the client header:
- sf-list of [sf-byte-sequences](https://www.rfc-editor.org/rfc/rfc8941.html#name-byte-sequences)
- sf-list of [sf-strings](https://www.rfc-editor.org/rfc/rfc8941.html#name-strings)
- sf-list of [sf-tokens](https://www.rfc-editor.org/rfc/rfc8941.html#name-tokens)
- [RFC 7639](https://datatracker.ietf.org/doc/rfc7639/)

Possible response syntax:
- one byte-sequence/string/token
- number indicating the index of the selected subprotocol

vasilvv:
Fixes #137

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -213,7 +213,9 @@ Both `WebTransport-Subprotocols-Available` and `WebTransport-Subprotocol` are
 Structured Fields {{!RFC8941}}. `WebTransport-Subprotocols-Available` is a List
 of Tokens, and `WebTransport-Subprotocol` is a Token. The token in the
 `WebTransport-Subprotocol` response header field MUST be one of the tokens
-listed in `WebTransport-Subprotocols-Available` of the request.
+listed in `WebTransport-Subprotocols-Available` of the request.  The semantics
+of individual token values is entirely up to the WebTransport resource in
+question.
 
 ## Limiting the Number of Simultaneous Sessions
 
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -206,8 +206,8 @@ this functionality.
 
 The user agent MAY include a `WebTransport-Subprotocols-Available` header field
 in the CONNECT request, enumerating the possible subprotocols; if received, the
-server MUST include a `WebTransport-Subprotocol` field in the 2xx response, or
-respond with an error code instead.
+server MUST include a `WebTransport-Subprotocol` field with the subprotocol
+chosen in the 2xx response, or respond with an error code instead.
 
 Both `WebTransport-Subprotocols-Available` and `WebTransport-Subprotocol` are
 Structured Fields {{!RFC8941}}. `WebTransport-Subprotocols-Available` is a List
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -215,7 +215,7 @@ of Tokens, and `WebTransport-Subprotocol` is a Token. The token in the
 `WebTransport-Subprotocol` response header field MUST be one of the tokens
 listed in `WebTransport-Subprotocols-Available` of the request.  The semantics
 of individual token values is determined by the WebTransport resource in
-question.
+question, and are not registered in IANA's "ALPN Protocol IDs" registry.
 
 ## Limiting the Number of Simultaneous Sessions
 
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -205,9 +205,11 @@ intent is to simplify porting pre-existing protocols that use QUIC and rely on
 this functionality.
 
 The user agent MAY include a `WebTransport-Subprotocols-Available` header field
-in the CONNECT request, enumerating the possible subprotocols; if received, the
-server MUST include a `WebTransport-Subprotocol` field with the subprotocol
-chosen in the 2xx response, or respond with an error code instead.
+in the CONNECT request, enumerating the possible subprotocols. If the server
+receives such a header, it MAY include a `WebTransport-Subprotocol` field in
+a successful (2xx) response. If it does, the server SHALL include a single
+subprotocol from the client's list in that field. Servers MAY reject the request
+if the client did not include a suitable subprotocol.
 
 Both `WebTransport-Subprotocols-Available` and `WebTransport-Subprotocol` are
 Structured Fields {{!RFC8941}}. `WebTransport-Subprotocols-Available` is a List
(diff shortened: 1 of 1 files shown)


draft-ietf-webtrans-http3.md
```diff
@@ -197,6 +197,24 @@ The `webtransport` HTTP Upgrade Token uses the Capsule Protocol as defined in
...
+## Subprotocol Negotiation
+
+WebTransport over HTTP/3 offers a subprotocol negotiation mechanism, similar to
+TLS Application-Layer Protocol Negotiation Extension (ALPN) {{?RFC7301}}; the
+intent is to simplify porting pre-existing protocols that use QUIC and rely on
+this functionality.
+
+The user agent MAY include a `WebTransport-Subprotocols-Available` header field
+in the CONNECT request, enumerating the possible subprotocols; if received, the
+server MUST include a `WebTransport-Subprotocol` field in the 2xx response, or
```
DavidSchinazi, line 209: What does the server put in this field?

draft-ietf-webtrans-http3.md
```diff
@@ -197,6 +197,24 @@ The `webtransport` HTTP Upgrade Token uses the Capsule Protocol as defined in
...
+respond with an error code instead.
+
+Both `WebTransport-Subprotocols-Available` and `WebTransport-Subprotocol` are
+Structured Fields {{!RFC8941}}. `WebTransport-Subprotocols-Available` is a List
+of Tokens, and `WebTransport-Subprotocol` is a Token. The token in the
+`WebTransport-Subprotocol` response header field MUST be one of the tokens
+listed in `WebTransport-Subprotocols-Available` of the request.
+
```
DavidSchinazi, line 217: Are these tokens registered with IANA? We should state the answer in the draft

DavidSchinazi (commented):
Overall SGTM

draft-ietf-webtrans-http3.md
vasilvv: Clarified.

draft-ietf-webtrans-http3.md
vasilvv: Clarified.

draft-ietf-webtrans-http3.md
```diff
@@ -197,6 +197,26 @@ The `webtransport` HTTP Upgrade Token uses the Capsule Protocol as defined in
...
+WebTransport over HTTP/3 offers a subprotocol negotiation mechanism, similar to
+TLS Application-Layer Protocol Negotiation Extension (ALPN) {{?RFC7301}}; the
+intent is to simplify porting pre-existing protocols that use QUIC and rely on
+this functionality.
+
+The user agent MAY include a `WebTransport-Subprotocols-Available` header field
+in the CONNECT request, enumerating the possible subprotocols; if received, the
+server MUST include a `WebTransport-Subprotocol` field with the subprotocol
+chosen in the 2xx response, or respond with an error code instead.
```
DavidSchinazi, line 210: This MUST makes this feature mandatory for servers (as opposed to optional for clients). I'd suggest keeping the optionality, similar to how ALPN works. For example:
```suggestion
in the CONNECT request, enumerating the possible subprotocols. If the server
receives such a header, it MAY include a `WebTransport-Subprotocol` field in
a successful (2xx) response. If it does, the server SHALL include a single
subprotocol from the client's list in that field. Servers MAY reject the request
if the client did not include a suitable subprotocol.
```

draft-ietf-webtrans-http3.md
```diff
@@ -197,6 +197,26 @@ The `webtransport` HTTP Upgrade Token uses the Capsule Protocol as defined in
...
+
+Both `WebTransport-Subprotocols-Available` and `WebTransport-Subprotocol` are
+Structured Fields {{!RFC8941}}. `WebTransport-Subprotocols-Available` is a List
+of Tokens, and `WebTransport-Subprotocol` is a Token. The token in the
+`WebTransport-Subprotocol` response header field MUST be one of the tokens
+listed in `WebTransport-Subprotocols-Available` of the request.  The semantics
+of individual token values is entirely up to the WebTransport resource in
+question.
```
DavidSchinazi, line 218:
```suggestion
question, and are not registered with IANA in the way that ALPN tokens are.
```

draft-ietf-webtrans-http3.md
vasilvv: Well, it's only mandatory if the client already expects the server to support it.

draft-ietf-webtrans-http3.md
vasilvv: This is technically not true. Assume I have a Foobar Protocol, and it has subprotocols `foo` and `bar`. I can create an IANA registry for subprotocols supported by the Foobar resources, and put `foo` and `bar` into it.

draft-ietf-webtrans-http3.md
```diff
@@ -197,6 +197,26 @@ The `webtransport` HTTP Upgrade Token uses the Capsule Protocol as defined in
...
+in the CONNECT request, enumerating the possible subprotocols; if received, the
+server MUST include a `WebTransport-Subprotocol` field with the subprotocol
+chosen in the 2xx response, or respond with an error code instead.
+
+Both `WebTransport-Subprotocols-Available` and `WebTransport-Subprotocol` are
+Structured Fields {{!RFC8941}}. `WebTransport-Subprotocols-Available` is a List
+of Tokens, and `WebTransport-Subprotocol` is a Token. The token in the
+`WebTransport-Subprotocol` response header field MUST be one of the tokens
+listed in `WebTransport-Subprotocols-Available` of the request.  The semantics
+of individual token values is entirely up to the WebTransport resource in
```
ekinnear, line 217:
```suggestion
of individual token values is entirely determined by the WebTransport resource in
```

ekinnear (approved):

draft-ietf-webtrans-http3.md
afrind: Feels like we ought to say something about the case where the client asked for "foo, bar" and the server says "200" but doesn't include the header.

draft-ietf-webtrans-http3.md
DavidSchinazi:
```suggestion
question, and are not registered in IANA's "ALPN Protocol IDs" registry.
```

DavidSchinazi (approved):

PR #148, opened 2024-02-28
Define key exporters

### LINKED ISSUE #116: Keying material exporters
Per w3c/webtransport#411, there is some interest in using [key exporters](https://datatracker.ietf.org/doc/rfc5705/) with WebTransport.  While in theory that could be purely an API concept, given that we do things like pooling (and given that we should register exporter labels with IANA), it probably would make sense to define the exact mechanism in this document.

I propose the following.  A WebTransport exporter takes `(label, context, length)` as input, similar to the TLS ones.  To derive a WebTransport exporter, derive a TLS key exporter with label `EXPORTER-WebTransport` and the context being the following:

```
WebTransport Exporter Context {
  WebTransport Session ID (64),
  WebTransport Application-Supplied Exporter Label Length (32),
  WebTransport Application-Supplied Exporter Label (..),
  WebTransport Application-Supplied Exporter Context (..)
}
```

This should help avoiding collisions with non-WebTransport exporters, as well as WebTransport exporters from different pooled sessions.  What do people think?

vasilvv:
Fixes #116

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -466,7 +466,7 @@ WebTransport over HTTP/3 supports the use of TLS keying material exporters
 WebTransport sessions, WebTransport defines its own mechanism for deriving a
 TLS exporter that separates keying material for different sessions.  If the
 user requests an exporter for a given WebTransport session with a specified
-label and context, the resulting exporter SHALL be a TLS exporter as defiend in
+label and context, the resulting exporter SHALL be a TLS exporter as defined in
 {{RFC5705, Section 4}} with the label set to "EXPORTER-WebTransport" and the
 context set to the serialization of the "WebTransport Exporter Context" struct
 as defined below.
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -481,8 +481,9 @@ WebTransport Exporter Context {
 }
 ~~~
 
-If the user does not supply any context, it is treated as the context
-with zero length.
+A TLS exporter API might permit the context field to be omitted. In this case,
+as with TLS 1.3, the WebTransport Application-Supplied Exporter Context
+becomes zero-length if omitted.
 
 
 # Session Termination
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -462,7 +462,7 @@ is expected to attempt to gracefully terminate the session as soon as possible.
 ## Use of Keying Material Exporters
 
 WebTransport over HTTP/3 supports the use of TLS keying material exporters
-{{!RFC5705}}.  Since the underlying QUIC connection may be shared by multiple
+{{!RFC8446, Section 7.5}}.  Since the underlying QUIC connection may be shared by multiple
 WebTransport sessions, WebTransport defines its own mechanism for deriving a
 TLS exporter that separates keying material for different sessions.  If the
 user requests an exporter for a given WebTransport session with a specified
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -467,7 +467,7 @@ WebTransport sessions, WebTransport defines its own mechanism for deriving a
 TLS exporter that separates keying material for different sessions.  If the
 user requests an exporter for a given WebTransport session with a specified
 label and context, the resulting exporter SHALL be a TLS exporter as defined in
-{{RFC8446, Section 7.5}} with the label set to "EXPORTER-WebTransport" and the
+{{!RFC8446, Section 7.5}} with the label set to "EXPORTER-WebTransport" and the
 context set to the serialization of the "WebTransport Exporter Context" struct
 as defined below.
 
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -476,7 +476,7 @@ WebTransport Exporter Context {
   WebTransport Session ID (64),
   WebTransport Application-Supplied Exporter Label Length (8),
   WebTransport Application-Supplied Exporter Label (..),
-  WebTransport Application-Supplied Exporter Context Length (32),
+  WebTransport Application-Supplied Exporter Context Length (8),
   WebTransport Application-Supplied Exporter Context (..)
 }
 ~~~
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -467,7 +467,7 @@ WebTransport sessions, WebTransport defines its own mechanism for deriving a
 TLS exporter that separates keying material for different sessions.  If the
 user requests an exporter for a given WebTransport session with a specified
 label and context, the resulting exporter SHALL be a TLS exporter as defined in
-{{RFC5705, Section 4}} with the label set to "EXPORTER-WebTransport" and the
+{{RFC8446, Section 7.5}} with the label set to "EXPORTER-WebTransport" and the
 context set to the serialization of the "WebTransport Exporter Context" struct
 as defined below.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-webtrans-http3.md
```diff
@@ -459,6 +459,31 @@ HTTP/3 GOAWAY frame, an endpoint MAY continue using the session and MAY open new
...
+user requests an exporter for a given WebTransport session with a specified
+label and context, the resulting exporter SHALL be a TLS exporter as defiend in
+{{RFC5705, Section 4}} with the label set to "EXPORTER-WebTransport" and the
+context set to the serialization of the "WebTransport Exporter Context" struct
+as defined below.
+
+~~~
+WebTransport Exporter Context {
+  WebTransport Session ID (64),
+  WebTransport Application-Supplied Exporter Label Length (32),
```
martinthomson, line 477: 32 bits is monstrously long.  TLS exporters only permit 0..255, so I would strongly suggest that you change these:

```suggestion
  WebTransport Application-Supplied Exporter Label Length (8),
```

You might also note that there is a further limitation on the total size imposed by TLS.

draft-ietf-webtrans-http3.md
```diff
@@ -459,6 +459,31 @@ HTTP/3 GOAWAY frame, an endpoint MAY continue using the session and MAY open new
...
 is expected to attempt to gracefully terminate the session as soon as possible.
 
+## Use of Keying Material Exporters
+
+WebTransport over HTTP/3 supports the use of TLS keying material exporters
+{{!RFC5705}}.  Since the underlying QUIC connection may be shared by multiple
+WebTransport sessions, WebTransport defines its own mechanism for deriving a
+TLS exporter that separates keying material for different sessions.  If the
+user requests an exporter for a given WebTransport session with a specified
+label and context, the resulting exporter SHALL be a TLS exporter as defiend in
```
martinthomson, line 469:
```suggestion
label and context, the resulting exporter SHALL be a TLS exporter as defined in
```

draft-ietf-webtrans-http3.md
```diff
@@ -459,6 +459,31 @@ HTTP/3 GOAWAY frame, an endpoint MAY continue using the session and MAY open new
...
+{{RFC5705, Section 4}} with the label set to "EXPORTER-WebTransport" and the
```
martinthomson, line 470:
```suggestion
{{RFC8446, Section 7.5}} with the label set to "EXPORTER-WebTransport" and the
```

draft-ietf-webtrans-http3.md
```diff
@@ -459,6 +459,31 @@ HTTP/3 GOAWAY frame, an endpoint MAY continue using the session and MAY open new
...
+  WebTransport Session ID (64),
+  WebTransport Application-Supplied Exporter Label Length (32),
+  WebTransport Application-Supplied Exporter Label (..),
+  WebTransport Application-Supplied Exporter Context Length (32),
+  WebTransport Application-Supplied Exporter Context (..)
+}
+~~~
+
+If the user does not supply any context, it is treated as the context
+with zero length.
```
martinthomson, line 485:
```suggestion
A TLS exporter API might permit the context field to be omitted. In this case,
as with TLS 1.3, the WebTransport Application-Supplied Exporter Context
becomes zero-length if omitted.
```

martinthomson (approved):

ekinnear (approved):

draft-ietf-webtrans-http3.md
```diff
@@ -459,6 +459,32 @@ HTTP/3 GOAWAY frame, an endpoint MAY continue using the session and MAY open new
...
+{{RFC8446, Section 7.5}} with the label set to "EXPORTER-WebTransport" and the
+context set to the serialization of the "WebTransport Exporter Context" struct
+as defined below.
+
+~~~
+WebTransport Exporter Context {
+  WebTransport Session ID (64),
+  WebTransport Application-Supplied Exporter Label Length (8),
+  WebTransport Application-Supplied Exporter Label (..),
+  WebTransport Application-Supplied Exporter Context Length (32),
```
martinthomson, line 479:
```suggestion
  WebTransport Application-Supplied Exporter Context Length (8),
```

draft-ietf-webtrans-http3.md
```diff
@@ -459,6 +459,32 @@ HTTP/3 GOAWAY frame, an endpoint MAY continue using the session and MAY open new
...
+  WebTransport Application-Supplied Exporter Context (..)
+}
+~~~
+
```
martinthomson, line 483:
```suggestion

Note that the TLS exporter API requires that labels are between 1 and 249 bytes in length and contexts are between 0 and 255 bytes.  Because this structure is included in the context field of a TLS exporter, the combined length of labels and contexts needs to be 245 bytes or shorter.

```

draft-ietf-webtrans-http3.md
```diff
@@ -459,6 +459,32 @@ HTTP/3 GOAWAY frame, an endpoint MAY continue using the session and MAY open new
...
+label and context, the resulting exporter SHALL be a TLS exporter as defined in
+{{RFC8446, Section 7.5}} with the label set to "EXPORTER-WebTransport" and the
+context set to the serialization of the "WebTransport Exporter Context" struct
+as defined below.
+
+~~~
+WebTransport Exporter Context {
+  WebTransport Session ID (64),
+  WebTransport Application-Supplied Exporter Label Length (8),
+  WebTransport Application-Supplied Exporter Label (..),
```
martinthomson, line 478:
```suggestion
  WebTransport Application-Supplied Exporter Label (8..),
```

It probably makes sense to set a minimum.

draft-ietf-webtrans-http3.md
martinthomson: Note: it is not sufficient to just merge suggestions.

vasilvv: Ah, I misread that as implying that only the label was subject to length restrictions.

vasilvv:
So, are those limits actually applicable?  At least in RFC8446, it looks like passes the hash of the context, and not the context itself.  I guess, we could limit ourselves to what would be allowed by TLS if we wanted them to be fully compatible.

DavidSchinazi:
Chair: discussed in the room at 119. No concerns from room. Plan is to give folks another week to review and then merge if there are no objections

draft-ietf-webtrans-http3.md
martinthomson: As Victor notes, this limit doesn't really need to be enforced due to the hashing limit.

draft-ietf-webtrans-http3.md
```diff
@@ -459,6 +459,32 @@ HTTP/3 GOAWAY frame, an endpoint MAY continue using the session and MAY open new
...
 
+## Use of Keying Material Exporters
+
+WebTransport over HTTP/3 supports the use of TLS keying material exporters
+{{!RFC5705}}.  Since the underlying QUIC connection may be shared by multiple
+WebTransport sessions, WebTransport defines its own mechanism for deriving a
+TLS exporter that separates keying material for different sessions.  If the
+user requests an exporter for a given WebTransport session with a specified
+label and context, the resulting exporter SHALL be a TLS exporter as defined in
+{{RFC8446, Section 7.5}} with the label set to "EXPORTER-WebTransport" and the
```
martinthomson, line 470:
```suggestion
{{!RFC8446, Section 7.5}} with the label set to "EXPORTER-WebTransport" and the
```

DavidSchinazi, line 470:
```suggestion
{{!RFC8446, Section 7.5}} with the label set to "EXPORTER-WebTransport" and the
```

draft-ietf-webtrans-http3.md
```diff
@@ -459,6 +459,32 @@ HTTP/3 GOAWAY frame, an endpoint MAY continue using the session and MAY open new
 streams.  The signal is intended for the application using WebTransport, which
 is expected to attempt to gracefully terminate the session as soon as possible.
 
+## Use of Keying Material Exporters
+
+WebTransport over HTTP/3 supports the use of TLS keying material exporters
+{{!RFC5705}}.  Since the underlying QUIC connection may be shared by multiple
```
martinthomson, line 465:
```suggestion
{{!RFC8446, Section 7.5}}.  Since the underlying QUIC connection may be shared by multiple
```

Issue #143, opened 2023-11-06
What does the server do if it receives a WebTransport request before the client's SETTINGS


DavidSchinazi:
As discussed in #139 and during the WG session at IETF 118, it is possible for the server to receives a WebTransport request before the client's SETTINGS. In that scenario, there are multiple options possible
* have the server respond to the webtransport request assuming that the client supports webtransport
* have the server buffer this request until it receives the client's SETTINGS
* have the server buffer all bidirectional stream data until it receives the client's SETTINGS

It was raised during the meeting that the latter options might make servers vulnerable to malicious clients that send many requests without ever sending SETTINGS.

We need to specify how the server behaves in this scenario

ekinnear:
Discussed in editor's meeting:

- Practically, if you support more than one WT version, you have to buffer CONNECT requests, until you know which version you're going to use (this arrives in the client's SETTINGS). The CONNECT itself might, in some WT version, contain information that depends on the WT version in use.
- When a new stream arrives, you have to parse far enough into it to determine that it's: 
    - a WebTransport stream
    - the WebTransport stream ID, so that you can tell if you have already received the CONNECT request/stream

Possible position: Therefore, you already have a buffering point on the server, so anything that we do around the client's SETTINGS arriving won't make a significant difference.

Server's that only support a single WT version can assume that it's that, or you can apply rules from all supported versions. If we're being careful in the future when we define new versions, we might take care to ensure that WT stream IDs and the surrounding items stay at the front so that you don't have to determine the version before you know what to do with the stream. Even if we do that, it doesn't tackle whether or not the CONNECT request has version-dependent headers.

One option is to say that we have an invariant across versions that all WT streams start with the type and the stream ID. We could potentially choose to include some indication of the version in the CONNECT request (either in :protocol or in a separate header, and then sort out all of the issues that arise if you send conflicting versions which should be possible but gross). All that would buy us is the ability to respond to incoming CONNECT requests _for WT versions that don't need non-default SETTINGS values_ and that also don't have a separate header in CONNECT requests that would carry such information.

DavidSchinazi:
[...]

Issue #141, opened 2023-11-05
don't require the client to send the WEBTRANSPORT_MAX_SESSIONS setting


marten-seemann:
There's no need for the client to send the WEBTRANSPORT_MAX_SESSIONS setting. The server doesn't need to know that a client supports WebTransport until it receives the first Extended CONNECT request that attempts to establish a WebTransport session.

Similar to #140, this requirement technically requires servers to buffer incoming requests until they have received the SETTINGS frame, if they want to check that the client actually sent the setting.

vasilvv:
It's a version negotiation mechanism. You can't figure out what version of WebTransport is being used until you see what version the client supports, and what version the server supports, since the version used is max(client versions ∩ server versions).

> Similar to https://github.com/ietf-wg-webtrans/draft-ietf-webtrans-http3/issues/140, this requirement technically requires servers to buffer incoming requests until they have received the SETTINGS frame, if they want to check that the client actually sent the setting.

Correct.

ekinnear:
For extra context, see also #125, #84, and especially #130 where it was proposed that the client not send the setting.

ekinnear:
If we can get to a place (and not run into issues with some of the other settings that get exchanged) that means we don't need to add any roundtrips before we can get going with WebTransport application data, I think we'll be very happy, so I would generally support anything that gets us closer to that goal.

vasilvv:
None of this causes any additional latency unless there is packet loss, and since you'd usually bundle your settings with ServerHello (for the server) or the CONNECT request (for the client), that should normally not matter for the client either.

marten-seemann:
> None of this causes any additional latency unless there is packet loss, and since you'd usually bundle your settings with ServerHello (for the server) or the CONNECT request (for the client), that should normally not matter for the client either.

I'm not worried about roundtrips here. The server sends SETTINGS in 0.5-RTT data, so they should be available as soon as the handshake completes. What I'm worried about are resource exhaustion attacks if the server has to buffer requests until it receives the client's SETTINGS. A malicious client could withhold sending of this frame for an arbitrarily long time.


> It's a version negotiation mechanism. You can't figure out what version of WebTransport is being used until you see what version the client supports, and what version the server supports, since the version used is max(client versions ∩ server versions).

In my HTTP/3 implementation, every HTTP request is handled immediately. There's no buffering of requests. I believe this is a reasonable design for defending against a variety of resource exhaustion attacks.
[...]

Issue #140, opened 2023-11-05
requirement on HTTP DATAGRAMs requires servers to buffer requests


marten-seemann:
In order to establish a WebTransport session, both client and server need to enable HTTP DATAGRAM support.

This is a problem on the server side, since SETTINGS are sent on a different stream, and the server might not have received the client's SETTINGS when it processes the Extended CONNECT request. If the server wants to verify the requirement from the draft, it will need to buffer the Extended CONNECT request until it has received the client's SETTINGS. Obviously, this is suboptimal, especially since a malicious client can delay sending of the SETTINGS.

A possible solution would be to revisit our previous decision to require datagram support for WebTransport. It would then be valid to have a WebTransport session that doesn't support datagrams.

vasilvv:
WebTransport already requires the server to not read any data from incoming bidirectional streams until it receives settings, because WebTransport changes the syntax of those in non-backwards-compatible manner.

marten-seemann:
That's not my understanding.

As soon as a server _sent_ the SETTINGS frame (containing all the extensions needed to use WebTransport), you need to be prepared to receive WebTransport streams. This shouldn't be a problem, since they start with the Signal Value (0x41), so the HTTP layer knows that it's a WebTransport stream. The WebTransport layer might need to buffer the stream if there's no session for that session ID, but there shouldn't be a situation where the HTTP layer needs to buffer anything.

kixelated:
Is this specific to datagrams? Wouldn't a pedantic WebTransport server also have to buffer the CONNECT and any streams until `SETTINGS_WEBTRANSPORT_MAX_SESSIONS` has been verified?

I agree that the race between SETTINGS and CONNECT is strange. The receipt of a CONNECT request with :protocol WebTransport implies that the client has sent the associated SETTINGS. But I think the draft is right:

> Similarly, the server MUST NOT process any incoming WebTransport requests until the client settings have been received, as the client may be using a version of the WebTransport extension that is different from the one used by the server.

This behavior is inevitable if the any SETTINGS changes the encoding/behavior of the CONNECT request. A smart server could defer verifying the SETTINGS frame but there's really no point.

marten-seemann:
> Is this specific to datagrams? Wouldn't a pedantic WebTransport server also have to buffer the CONNECT and any streams until `SETTINGS_WEBTRANSPORT_MAX_SESSIONS` has been verified?

[...]

PR #149, opened 2024-03-04
Adjust SETTINGS, buffering requirements, waiting for session establishment

### LINKED ISSUE #135: Should clients wait for SETTINGS?
HTTP syntax is extended by client-initiated bidirectional streams, which is why the settings for enabling WebTransport matter more than it would otherwise. This raises the question if we should have included a "fake" very high length for the "frame that runs to the end of the stream" so that a parser that's doing the wrong thing won't see the session ID and try to parse out that number of bytes.

The question is if that would matter? Would it help? Is it worth the cost (9 bytes per stream)?

The scenario: 
1. Client tries to do WT, thinking that the server does it, but it does not. 
2. Server sees a frame of a type that it doesn't recognize, so tries to read the length to skip that frame.
3. The length is actually the WT Session ID, so now the server reads that far through the byte stream and attempts to parse the next bytes as the beginning of an HTTP/3 frame. 
4. You now have the ability to write whatever HTTP/3 frames you want to the server, but the client has no idea what you're saying. 

- Hopefully, the client doesn't allow you to talk WT to a server that doesn't actually talk WT
- If you had a client that was either malicious or buggy, you could try to send these frames

Impact: 
- What sort of information could you get by sending custom HTTP/3 frames to the server? 
- This would be a CORS bypass in a browser
- Allows the attacker to access secrets in the dynamic table
     - If the connection is shared with other HTTP requests, that's especially concerning
     - You can put indices that will extract headers and then talk to an echo endpoint on the same server that will reflect them back to you
    - You can also use these headers, for example authorization

Ways to fix this: 
a. Send a sentinel frame (say, empty SETTINGS, 4 and 0) that means nothing ahead of the WT frames. WT servers will know and discard. Other implementations will fail and tear down the whole connection. 
b. Considered using max length in the WT frame, but the failure mode is less good (if someone tries to allocate that much memory, we have issues, plus it's not great to have things stall).
c. We should write down that client MUST wait until settings (this exists, but we should add it to security considerations)

Overall:
Don't offer a WT API that allows people to write arbitrary bytes to various streams.

ekinnear:
Adjust SETTINGS, buffering requirements, waiting for session establishment

As discussed: 
- Client MUST wait for server settings
- Endpoints MUST send SETTINGS_H3_DATAGRAM (however servers that want to validate this MUST need to take into account the fact that the request might arrive before the client's SETTINGS)
- Client no longer sends the WEBTRANSPORT_MAX_SESSIONS setting
- Add a "Considerations for future versions of WebTransport" section that explain that if a future version of WebTransport changes the syntax of the request, it'll need to change the Upgrade Token. Similarly, changes to stream formats will require changes to the Unidirectional Stream Type and Bidirectional Stream Signal Value.

Fixes #135, #140, #141, and #143

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -81,10 +81,10 @@ can be accessed via an HTTP/3 server.
 WebTransport servers in general are identified by a pair of authority value and
 path value (defined in {{!RFC3986}} Sections 3.2 and 3.3 correspondingly).
 
-When an HTTP/3 connection is established, both the client and server have to
-send a SETTINGS_WEBTRANSPORT_MAX_SESSIONS setting in order to indicate that
-they both support WebTransport over HTTP/3.  This process also negotiates the
-use of additional HTTP/3 extensions.
+When an HTTP/3 connection is established, the server sends a
+SETTINGS_WEBTRANSPORT_MAX_SESSIONS setting in order to indicate support for
+WebTransport over HTTP/3.  This process also negotiates the use of additional
+HTTP/3 extensions.
 
 WebTransport sessions are initiated inside a given HTTP/3 connection by the
 client, who sends an extended CONNECT request {{!RFC8441}}.  If the server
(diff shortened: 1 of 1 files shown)


ekinnear:
(Double checked, merging now so we can discuss in more detail at 119, happy to spin additional updates as needed!)

PR #151, opened 2024-03-17
Make WebTransport resets reliable

### LINKED ISSUE #77: Stream resets can leave streams half-closed indefinitely
Consider the following situation:
1. A opens a bidirectional WebTransport stream to B, sends some application data, and then decides to abort data transmission (i.e. it sends a RESET_STREAM frame)
2. B receives the RESET_STREAM frame before it receives the application data, and surfaces the stream reset to the WebTransport application before it had the chance to read the Session ID. Note that there's no way around this: the STREAM frames might have been lost, and A won't retransmit them since the stream was reset.

We're now in the situation where one side of the stream is reset, but the other side of the stream is still open. Unless _someone_ explicitly closes / resets the other side of the stream, it will stay this way indefinitely.

I can see two ways forward here:
1. Require A to send a STOP_SENDING frame as well. This is a kind of sad solution, as this would be a significant deviation from the QUIC stream state machine.
2. Require B to reset streams iff it processes B's stream reset before reading the session ID. We'd have to define the error code to use for that.

vasilvv:
Fixes #77

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -146,8 +146,7 @@ of !QUIC-DATAGRAM=RFC9221}}).
 
 WebTransport over HTTP/3 relies on the RESET_STREAM_AT frame defined in
 {{!RESET-STREAM-AT=I-D.ietf-quic-reliable-stream-reset}}.  To indicate support,
-both the client and the server MUST send a reliable_stream_reset transport
-parameter as defined in {{Section 3 of RESET-STREAM-AT}}.
+both the client and the server MUST enable the extension as described in {{Section 3 of RESET-STREAM-AT}}.
 
 ## Extended CONNECT in HTTP/3
 
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -146,7 +146,8 @@ of !QUIC-DATAGRAM=RFC9221}}).
 
 WebTransport over HTTP/3 relies on the RESET_STREAM_AT frame defined in
 {{!RESET-STREAM-AT=I-D.ietf-quic-reliable-stream-reset}}.  To indicate support,
-both the client and the server MUST enable the extension as described in {{Section 3 of RESET-STREAM-AT}}.
+both the client and the server MUST enable the extension as described 
+in {{Section 3 of RESET-STREAM-AT}}.
 
 ## Extended CONNECT in HTTP/3
 
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -144,6 +144,11 @@ support, both the client and the server MUST send a max_datagram_frame_size
 transport parameter with a value greater than 0 (see {{Section 3
 of !QUIC-DATAGRAM=RFC9221}}).
 
+WebTransport over HTTP/3 relies on the RESET_STREAM_AT frame defined in
+{{!RESET-STREAM-AT=I-D.ietf-quic-reliable-stream-reset}}.  To indicate support,
+both the client and the server MUST send a reliable_stream_reset transport
+parameter as defined in {{Section 3 of RESET-STREAM-AT}}.
+
 ## Extended CONNECT in HTTP/3
 
 {{!RFC8441}} defines an extended CONNECT method in Section 4, enabled by the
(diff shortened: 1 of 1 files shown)


draft-ietf-webtrans-http3.md
```diff
@@ -144,6 +144,11 @@ support, both the client and the server MUST send a max_datagram_frame_size
 transport parameter with a value greater than 0 (see {{Section 3
 of !QUIC-DATAGRAM=RFC9221}}).
 
+WebTransport over HTTP/3 relies on the RESET_STREAM_AT frame defined in
+{{!RESET-STREAM-AT=I-D.ietf-quic-reliable-stream-reset}}.  To indicate support,
+both the client and the server MUST send a reliable_stream_reset transport
+parameter as defined in {{Section 3 of RESET-STREAM-AT}}.
```
marten-seemann, line 150:
```suggestion
both the client and the server MUST enable the extension as described in {{Section 3 of RESET-STREAM-AT}}.
```

draft-ietf-webtrans-http3.md
```diff
@@ -396,11 +401,10 @@ application errors and HTTP/3 error codes" }
...
 beginning of the stream; resetting a stream may result in that data being
-discarded.  Because of that, WebTransport application error codes are best
-effort, as the WebTransport stack is not always capable of associating the
-reset code with a session.  The only exception is the situation where there is
-only one session on a given HTTP/3 connection, and no intermediaries between
-the client and the server.
+discarded.  Because of that, WebTransport implementations MUST use the
+RESET_STREAM_AT frame {{RESET-STREAM-AT}} when resetting WebTransport data
+streams to ensure that the stream header containing the associated WebTransport
+session ID is always delivered.
```
marten-seemann, line 407: Suggestion (GitHub doesn't allow me to mark this a suggestion since it covers unchanged lines):
> WebTransport data streams are associated with sessions through a header at the
beginning of the stream; resetting a stream might result in that data being
discarded when using a RESET_STREAM frame.
> To prevent this, WebTransport implementations MUST use the RESET_STREAM_AT frame {{RESET-STREAM-AT}} with a Reliable Size set to at least the size of the header when resetting WebTransport data. This ensure that the stream header containing the associated WebTransport session ID is always delivered.

draft-ietf-webtrans-http3.md
```diff
@@ -396,11 +401,10 @@ application errors and HTTP/3 error codes" }
 
 WebTransport data streams are associated with sessions through a header at the
 beginning of the stream; resetting a stream may result in that data being
-discarded.  Because of that, WebTransport application error codes are best
-effort, as the WebTransport stack is not always capable of associating the
-reset code with a session.  The only exception is the situation where there is
-only one session on a given HTTP/3 connection, and no intermediaries between
-the client and the server.
+discarded.  Because of that, WebTransport implementations MUST use the
+RESET_STREAM_AT frame {{RESET-STREAM-AT}} when resetting WebTransport data
```
martinthomson, line 405: 'MUST use' ... how?

martinthomson (commented):
Take Marten's suggestions first, but I think that you need to elevate rst-str-at to the same level as h3-dg.

draft-ietf-webtrans-http3.md
vasilvv: Thanks! Adopted this with a few modifications.

ekinnear (approved):

aboba (approved):

draft-ietf-webtrans-http3.md
```diff
@@ -144,6 +144,10 @@ support, both the client and the server MUST send a max_datagram_frame_size
 transport parameter with a value greater than 0 (see {{Section 3
 of !QUIC-DATAGRAM=RFC9221}}).
 
+WebTransport over HTTP/3 relies on the RESET_STREAM_AT frame defined in
+{{!RESET-STREAM-AT=I-D.ietf-quic-reliable-stream-reset}}.  To indicate support,
+both the client and the server MUST enable the extension as described in {{Section 3 of RESET-STREAM-AT}}.
```
ekinnear, line 149:
```suggestion
both the client and the server MUST enable the extension as described 
in {{Section 3 of RESET-STREAM-AT}}.
```

PR #154, opened 2024-03-24
don't require the client to send SETTINGS_ENABLE_CONNECT_PROTOCOL

### LINKED ISSUE #153: Does the client really need to send SETTINGS_ENABLE_CONNECT_PROTOCOL?
With #149, the draft now says:
> To use WebTransport over HTTP/3, clients MUST send the SETTINGS_ENABLE_CONNECT_PROTOCOL setting with a value of "1".

I don't understand why this is necessary. SETTINGS_ENABLE_CONNECT_PROTOCOL allows the client to send Extended CONNECT requests. Why does the client need to declare that it supports the Extended CONNECT protocol if it can just send an Extended CONNECT request?

My understanding of RFC 8441 is that it's not required:
> Upon receipt of SETTINGS_ENABLE_CONNECT_PROTOCOL with a value of 1, a client MAY use the Extended CONNECT as defined in this document when creating new streams.  Receipt of this parameter by a server does not have any impact.

---

Note the similarities to a similar discussion we had about SETTINGS_WEBTRANSPORT_MAX_SESSIONS, where we concluded that it's not necessary for the client to send this setting.

marten-seemann:
Fixes #153.

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -150,9 +150,7 @@ of !QUIC-DATAGRAM=RFC9221}}).
 SETTINGS_ENABLE_CONNECT_PROTOCOL setting.  That setting is defined for HTTP/3
 by {{!RFC9220}}.  A server supporting WebTransport over HTTP/3 MUST send both
 the SETTINGS_WEBTRANSPORT_MAX_SESSIONS setting with a value greater than "0"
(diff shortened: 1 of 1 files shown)


DavidSchinazi (approved):

martinthomson (approved):

ekinnear (approved):
Thank you!

Issue #155, opened 2024-03-24
What happens if the client doesn't enable HTTP datagrams?


marten-seemann:
With #149, the server is not required to wait for the client's HTTP/3 SETTINGS anymore:
> Servers should also note that CONNECT requests to establish new WebTransport sessions, in addition to other messages, may arrive before this SETTING is received (see [Section 4.5](https://ietf-wg-webtrans.github.io/draft-ietf-webtrans-http3/draft-ietf-webtrans-http3.html#buffering-incoming)).

We should specify what happens if the client sends invalid SETTINGS (e.g. by not enabling HTTP datagrams). There are two options:
1. Kill the underlying QUIC connection. We should probably define an HTTP/3 error code for this.
2. Kill the WebTransport session(s). It's not clear how to do this, since the CLOSE_WEBTRANSPORT_SESSION capsule only carries an application error, and there's currently no way to send a (web-) transport error. If we decide to do flow control in WebTransport, this seems like something we'll need to add anyway though.

PR #165, opened 2024-06-11
WebTransport requests are malformed if the settings are wrong

afrind:
Fixes: #155

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -214,7 +214,8 @@ transport parameter with a value greater than 0 (see {{Section 3
 of !QUIC-DATAGRAM=RFC9221}}).
 
 Any WebTransport requests sent by the client without enabling QUIC and HTTP
-datagrams MUST be treated as malformed by the server.
+datagrams MUST be treated as malformed by the server,
+as defined in {{Section 4.1.2 of HTTP3}}.
 
 WebTransport over HTTP/3 relies on the RESET_STREAM_AT frame defined in
 {{!RESET-STREAM-AT=I-D.ietf-quic-reliable-stream-reset}}.  To indicate support,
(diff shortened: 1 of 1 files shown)


draft-ietf-webtrans-http3.md
```diff
@@ -213,9 +213,12 @@ support, both the client and the server MUST send a max_datagram_frame_size
 transport parameter with a value greater than 0 (see {{Section 3
 of !QUIC-DATAGRAM=RFC9221}}).
 
+Any WebTransport requests sent by the client without enabling QUIC and HTTP
+datagrams MUST be treated as malformed by the server.
```
marten-seemann, line 217: Does this mean the server should send a 400 or a H3_MESSAGE_ERROR?

afrind: Here's what 9114 says:

> Malformed requests or responses that are detected MUST be treated as a [stream error](https://datatracker.ietf.org/doc/html/rfc9114#errors) of type [H3_MESSAGE_ERROR](https://datatracker.ietf.org/doc/html/rfc9114#H3_MESSAGE_ERROR).

> For malformed requests, a server MAY send an HTTP response indicating the error prior to closing or resetting the stream.

So MUST reset and MAY send a response first (hope it's a reliable reset!).  I like this since it handles the case where the server might have already assumed the settings would be correct, and sent a 200, and detected malformed-ness later.

vasilvv: I like this.  Could you add a direct citation into the spec text?

DavidSchinazi:
```suggestion
datagrams MUST be treated as malformed by the server,
as defined in {{Section 4.1.2 of HTTP3}}.
```

draft-ietf-webtrans-http3.md
```diff
@@ -213,9 +213,13 @@ support, both the client and the server MUST send a max_datagram_frame_size
 transport parameter with a value greater than 0 (see {{Section 3
 of !QUIC-DATAGRAM=RFC9221}}).
 
+Any WebTransport requests sent by the client without enabling QUIC and HTTP
+datagrams MUST be treated as malformed by the server,
+as defined in {{Section 4.1.2 of HTTP3}}.
```
vasilvv, line 218:
```suggestion
as described in {{Section 4.1.2 of HTTP3}}.
```

vasilvv (approved):

PR #166, opened 2024-06-21
Add initial flow control text and capsules from HTTP/2

### LINKED ISSUE #85: MAX_STREAMS and flow control

ekinnear:
Add initial flow control text and capsules from HTTP/2 and draft-thomson-webtrans-session-limit-01. Happy to re-re-arrange sections as necessary.

Negotiation of flow control is coming under the separate issue to make things a little easier to keep track of and review.

Closes #85.


draft-ietf-webtrans-http3.md
```diff
@@ -568,6 +543,298 @@ A TLS exporter API might permit the context field to be omitted. In this case,
...
+
+This document defines a SETTINGS_WEBTRANSPORT_MAX_SESSIONS parameter that allows
+the server to limit the maximum number of concurrent WebTransport sessions on a
+single HTTP/3 connection.  The client MUST NOT open more sessions than
+indicated in the server SETTINGS parameters.  The server MUST NOT close the
+connection if the client opens sessions exceeding this limit, as the client and
+the server do not have a consistent view of how many sessions are open due to
+the asynchronous nature of the protocol; instead, it MUST reset all of the
+CONNECT streams it is not willing to process with the `HTTP_REQUEST_REJECTED`
+status defined in [HTTP3].
```
martinthomson, line 578:
```suggestion
status defined in {{HTTP3}}.
```

As a general rule, I recommend {{}} over [] because the latter will silently fail whereas the former does not.

draft-ietf-webtrans-http3.md
```diff
@@ -568,6 +543,298 @@ A TLS exporter API might permit the context field to be omitted. In this case,
...
+created within the session if both the stream- and the connection-level limit
+permit; see {{Section 4.6 of !RFC9000}} for details on how QUIC stream limits
+are applied.
+
+Unlike the WT_MAX_STREAMS capsule or the QUIC MAX_STREAMS frame, there is no
+simple relationship between the value in this frame and stream IDs in QUIC
+STREAM frames.  This especially applies if there are other users of streams on
+the connection.
+
+The WT_STREAMS_BLOCKED capsule ({{WT_STREAMS_BLOCKED}}) is sent to indicate that
```
martinthomson, line 612:
```suggestion
The WT_STREAMS_BLOCKED capsule ({{WT_STREAMS_BLOCKED}}) can be sent to indicate that
```

draft-ietf-webtrans-http3.md
```diff
@@ -568,6 +543,298 @@ A TLS exporter API might permit the context field to be omitted. In this case,
...
+(see {{unidirectional-streams}} and {{bidirectional-streams}}).  The stream
+header is excluded from this limit so that this limit does not prevent the
+sending of information that is essential in linking new streams to a specific
+WebTransport session.
+
+Implementing WT_MAX_DATA requires that the QUIC stack provide the WebTransport
+implementation with information about the final size of streams; see {{Section
+4.5 of !RFC9000}}.
+
+The WT_DATA_BLOCKED capsule ({{WT_DATA_BLOCKED}}) is sent to indicate that an
```
martinthomson, line 630:
```suggestion
The WT_DATA_BLOCKED capsule ({{WT_DATA_BLOCKED}}) can be sent to indicate that an
```

draft-ietf-webtrans-http3.md
```diff
@@ -568,6 +543,298 @@ A TLS exporter API might permit the context field to be omitted. In this case,
...
+data that can be sent within a WebTransport session.  This limit counts all data
+that is sent on streams of the corresponding type, excluding the stream header
+(see {{unidirectional-streams}} and {{bidirectional-streams}}).  The stream
+header is excluded from this limit so that this limit does not prevent the
+sending of information that is essential in linking new streams to a specific
+WebTransport session.
+
+Implementing WT_MAX_DATA requires that the QUIC stack provide the WebTransport
+implementation with information about the final size of streams; see {{Section
+4.5 of !RFC9000}}.
```
martinthomson, line 628: Is it worth highlighting this as a difference?  With HTTP/2 you don't need this, but HTTP/3 means that you are creating streams all over and using final size.  We might also note that this is independent of any RESET_STREAM_AT functionality.

draft-ietf-webtrans-http3.md
```diff
@@ -568,6 +543,298 @@ A TLS exporter API might permit the context field to be omitted. In this case,
...
+responsible for storing any data for which they advertise flow control credit
+if that data cannot be immediately forwarded to the next hop.
+
+In practice, an intermediary that translates flow control signals between
+similar WebTransport protocols, such as between two HTTP/3 connections, can
+often simply reexpress the same limits received on one connection directly on
+the other connection.
+
+An intermediary that does not want to be responsible for storing data that
+cannot be immediately sent on its translated connection would ensure that it
```
martinthomson, line 663:
```suggestion
cannot be immediately sent on its translated connection can ensure that it
```

martinthomson (approved):
Thanks for doing the heavy lifting @ekinnear.  I don't have any major comments.  You got the main ones, like intermediary responsibilities.

draft-ietf-webtrans-http3.md
ekinnear: When you say highlighting this as a difference, what're you thinking? Like adding additional text to describe how this works here as opposed to H2? (I'd imagine we'd add that to H2 where we refer to this capsule?)

draft-ietf-webtrans-http3.md
```diff
@@ -568,6 +543,298 @@ A TLS exporter API might permit the context field to be omitted. In this case,
...
+servers that wish to limit the rate of incoming requests on any particular
+session have alternative mechanisms:
+
+* The `HTTP_REQUEST_REJECTED` error code defined in [HTTP3] indicates to the
+  receiving HTTP/3 stack that the request was not processed in any way.
+* HTTP status code 429 indicates that the request was rejected due to rate
+  limiting {{!RFC6585}}.  Unlike the previous method, this signal is directly
+  propagated to the application.
+
+## Limiting the Number of Streams Within a Session {#flow-control-limit-streams}
```
vasilvv, line 593: We should probably point out somewhere inside this section that this mechanism only works when you have reliable resets for stream headers.

draft-ietf-webtrans-http3.md
```diff
@@ -568,6 +543,298 @@ A TLS exporter API might permit the context field to be omitted. In this case,
...
+data that can be sent within a WebTransport session.  This limit counts all data
+that is sent on streams of the corresponding type, excluding the stream header
+(see {{unidirectional-streams}} and {{bidirectional-streams}}).  The stream
+header is excluded from this limit so that this limit does not prevent the
+sending of information that is essential in linking new streams to a specific
+WebTransport session.
+
+Implementing WT_MAX_DATA requires that the QUIC stack provide the WebTransport
+implementation with information about the final size of streams; see {{Section
+4.5 of !RFC9000}}.
```
vasilvv, line 628: I didn't realize this.  Actually, the more I think about WT_MAX_DATA, the more concerning I find it from the implementation perspective.

So, to implement this mechanism, the WebTransport stack has to
1. Do its own accounting of how many bytes were written for the session in total.
2. Block whenever the application tries to write in excess.
3. Record all blocked streams.
4. Wake those streams up when the credit has been received, in the priority order.

While none of those steps seems to be individually challenging, that's a lot of machinery, leaving implementers quite a bit in terms of room to screw things up.  It is also really tempting for implementers to merge QUIC-level flow control blocks with the WebTransport-level ones, probably leading to large cross-layer mess.

vasilvv (commented):
Overall this seems reasonable, but I am still concerned about WT_MAX_DATA.

draft-ietf-webtrans-http3.md
```diff
@@ -568,6 +543,298 @@ A TLS exporter API might permit the context field to be omitted. In this case,
...
+connection if the client opens sessions exceeding this limit, as the client and
+the server do not have a consistent view of how many sessions are open due to
+the asynchronous nature of the protocol; instead, it MUST reset all of the
+CONNECT streams it is not willing to process with the `HTTP_REQUEST_REJECTED`
+status defined in {{HTTP3}}.
+
+Just like other HTTP requests, WebTransport sessions, and data sent on those
+sessions, are counted against flow control limits.  This document does not
+introduce additional mechanisms for endpoints to limit the relative amount of
+flow control credit consumed by different WebTransport sessions, however
```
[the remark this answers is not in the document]
ekinnear: Removed in 78d2456

draft-ietf-webtrans-http3.md
ekinnear: Added some initial text for this in 78d2456

draft-ietf-webtrans-http3.md
ekinnear: Hmm, I'm not sure how different this is from what an application needs to do in order to correctly provide back pressure for any connection that is multiplexing multiple independent streams. 

In that case, an application needs to: 
- Respond to back pressure signals from the underlying transport (TLS/TCP or QUIC), which then requires
- Blocking when the application tries to write in excess
- Keep track of which streams were blocked
- Tell those streams when they can write again / backpressure has been lifted. Ideally, in priority order.

Handling the fact that you may be restricted in how much you can send is something that everyone doing this muxing/demuxing needs to do, especially given priority -- priority only matters when there's a restriction on what we can send and we need to choose. If an implementation didn't already need to deal with this, we wouldn't need priorities at all.

Based on that, it seems like the main delta here is really just doing your own accounting of what that limit is vs. having someone below you tell you that you've run out.

ekinnear:
@vasilvv updated, please take a look!

draft-ietf-webtrans-http3.md
```diff
@@ -571,6 +546,288 @@ A TLS exporter API might permit the context field to be omitted. In this case,
...
+
+   Maximum Streams:
+   : A count of the cumulative number of streams of the corresponding type that
+     can be opened over the lifetime of the connection. This value cannot
+     exceed 2<sup>60</sup>, as it is not possible to encode stream IDs larger
+     than 2<sup>62</sup>-1.
+
+An endpoint MUST NOT open more streams than permitted by the current stream
+limit set by its peer.  For instance, a server that receives a unidirectional
+stream limit of 3 is permitted to open streams 3, 7, and 11, but not stream
```
afrind, line 710: Just talking out loud to remind myself - 

WT doesn't need logic about "implicitly" opening streams that have not arrived with lower stream numbers, like QUIC does, because there's no guarantee those quic streams are bound for this session.  A receiver just maintains a counter of how many streams have been opened and if that counter ever exceeds the last WT_MAX_STREAMS value then it's an error.

draft-ietf-webtrans-http3.md
```diff
@@ -571,6 +546,288 @@ A TLS exporter API might permit the context field to be omitted. In this case,
...
+  Maximum Data (i),
+}
+~~~
+{: #fig-wt_max_data title="WT_MAX_DATA Capsule Format"}
+
+WT_MAX_DATA capsules contain the following field:
+
+   Maximum Data:
+   : A variable-length integer indicating the maximum amount of data that can be
+     sent on the entire connection, in units of bytes.
```
afrind, line 788: Do we mean session and not connection?

draft-ietf-webtrans-http3.md
```diff
@@ -571,6 +546,288 @@ A TLS exporter API might permit the context field to be omitted. In this case,
...
+  Maximum Streams (i),
+}
+~~~
+{: #fig-wt_max_streams title="WT_MAX_STREAMS Capsule Format"}
+
+WT_MAX_STREAMS capsules contain the following field:
+
+   Maximum Streams:
+   : A count of the cumulative number of streams of the corresponding type that
+     can be opened over the lifetime of the connection. This value cannot
```
afrind, line 704: s/connection/session/ ?

draft-ietf-webtrans-http3.md
afrind: It might help the average reader to explain why they need final size and how the webtransport implementation should use it.

ekinnear: Added a small note about the stream header exclusion and we'll go through in more detail about final size differences across layers in our upcoming meeting.

draft-ietf-webtrans-http3.md
ekinnear: Good catch, updated, thanks!

draft-ietf-webtrans-http3.md
ekinnear: Updated!

afrind (approved):

PR #169, opened 2024-07-18
Fix some typos

MarcoPolo:
I'm not 100% sure about the change from:

"If at any point a session ID is received that cannot a valid ID for a"
to
"If at any point a session ID is received that cannot be a valid ID for a"

So please double check that logic. I went with the smallest hamming distance.

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -169,7 +169,7 @@ following mechanisms:
 * A client can create a bidirectional stream and transfer its ownership to
   WebTransport by providing a special signal in the first bytes.
 * A server can create a bidirectional stream and transfer its ownership to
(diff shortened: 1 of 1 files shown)


draft-ietf-webtrans-http3.md
```diff
@@ -592,7 +592,7 @@ CLOSE_WEBTRANSPORT_SESSION Capsule {
   Type (i) = CLOSE_WEBTRANSPORT_SESSION,
   Length (i),
   Application Error Code (32),
-  Application Error Message (..8192),
+  Application Error Message (..1024),
```
marten-seemann, line 595: The length is measured in bits, not bytes.

marten-seemann (changes requested):
The other changes lgtm.

DavidSchinazi (approved):

PR #181, opened 2025-03-25
Register HTTP fields

### LINKED ISSUE #164: WT-Available-Protocols IANA registration
Nothing in the IANA considerations section for this.  We should do that.

martinthomson:
Closes #164.

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -278,7 +278,7 @@ The `webtransport` HTTP Upgrade Token uses the Capsule Protocol as defined in
 HTTP-DATAGRAM}} is not required by WebTransport and can safely be ignored by WebTransport
 endpoints.
 
-## Application Protocol Negotiation
+## Application Protocol Negotiation {#protocol-negotiation}
 
 WebTransport over HTTP/3 offers a protocol negotiation mechanism, similar to
 TLS Application-Layer Protocol Negotiation Extension (ALPN) {{?RFC7301}}; the
@@ -295,9 +295,11 @@ reject the request if the client did not include a suitable protocol.
 Both `WT-Available-Protocols` and `WT-Protocol` are Structured Fields
 {{!RFC8941}}. `WT-Available-Protocols` is a List of Tokens, and `WT-Protocol` is
 a Token. The token in the `WT-Protocol` response header field MUST be one of
-the tokens listed in `WT-Available-Protocols` of the request.  The semantics of
-individual token values is determined by the WebTransport resource in question
-and are not registered in IANA's "ALPN Protocol IDs" registry.
+the tokens listed in `WT-Available-Protocols` of the request.
+
+The semantics of individual token values used in `WT-Available-Protocols` and
+`WT-Protocol` are determined by the WebTransport resource in question and are
+not registered in IANA's "ALPN Protocol IDs" registry.
 
 ## Prioritization
 
@@ -1195,7 +1197,7 @@ Specification:
 The following entries are added to the "HTTP Capsule Types" registry established
 by {{HTTP-DATAGRAM}}:
 
-The `CLOSE_WEBTRANSPORT_SESSION` capsule.
+The `CLOSE_WEBTRANSPORT_SESSION` capsule:
 
 Value:
 : 0x2843
@@ -1219,7 +1221,7 @@ Notes:
 : None
 {: spacing="compact"}
 
-The `DRAIN_WEBTRANSPORT_SESSION` capsule.
+The `DRAIN_WEBTRANSPORT_SESSION` capsule:
 
 Value:
 : 0x78ae
@@ -1243,7 +1245,7 @@ Notes:
 : None
 {: spacing="compact"}
 
-The `WT_MAX_STREAMS` capsule.
+The `WT_MAX_STREAMS` capsule:
 
 Value:
 : 0x190B4D3F..0x190B4D40
@@ -1267,7 +1269,7 @@ Notes:
 : None
 {: spacing="compact"}
 
-The `WT_STREAMS_BLOCKED` capsule.
+The `WT_STREAMS_BLOCKED` capsule:
 
 Value:
 : 0x190B4D43..0x190B4D44
@@ -1291,7 +1293,7 @@ Notes:
 : None
 {: spacing="compact"}
 
-The `WT_MAX_DATA` capsule.
+The `WT_MAX_DATA` capsule:
 
 Value:
 : 0x190B4D3D
@@ -1315,7 +1317,7 @@ Notes:
 : None
 {: spacing="compact"}
 
-The `WT_DATA_BLOCKED` capsule.
+The `WT_DATA_BLOCKED` capsule:
 
 Value:
 : 0x190B4D41
@@ -1338,6 +1340,50 @@ Contact:
 Notes:
 : None
 {: spacing="compact"}
+
+
+## Protocol Negotiation HTTP Header Fields {#iana-http}
+
+The following HTTP header fields are used for negotiating a protocol
+({{protocol-negotiation}}.  These are added to the "HTTP Field Name" registry
+established in {{Section 18.4 of HTTP}}:
+
+The `WT-Available-Protocols` field:
+
+Field Name:
+: WT-Available-Protocols
+
+Status:
+: permanent
+
+Structured Type:
+: List of Token
+
+Reference:
+: {{protocol-negotiation}}
+
+Comments:
+: None
+{: spacing="compact"}
+
+The `WT-Protocol` field:
+
+Field Name:
+: WT-Protocol
+
+Status:
+: permanent
+
+Structured Type:
+: Token
+
+Reference:
+: {{protocol-negotiation}}
+
+Comments:
+: None
+{: spacing="compact"}
+
 
 --- back
 


afrind (approved):

DavidSchinazi (approved):

draft-ietf-webtrans-http3.md
```diff
@@ -293,11 +293,19 @@ SHALL include a single choice from the client's list in that field. Servers MAY
...
-a Token. The token in the `WT-Protocol` response header field MUST be one of
-the tokens listed in `WT-Available-Protocols` of the request.  The semantics of
-individual token values is determined by the WebTransport resource in question
-and are not registered in IANA's "ALPN Protocol IDs" registry.
+{{!RFC8941}}. `WT-Available-Protocols` is a List, with the only valid value type
+being either String or Token.  `WT-Protocol` is defined as an item with the only
+valid type for values being String or Token. Any value type other than String or
+Token MUST be ignored.  String or Token values can be used interchangeably;
+decoded values are used to identify protocols. No semantics are defined for
+parameters on either field; parameters MUST be ignored.
```
vasilvv, line 301: Is this required for some reasons that I am not structured-headers enough to understand?

martinthomson: Structured fields is a bit loose.  There are assumed norms, but you can't just say "this is an item" without defining anything.  [Section 2 of RFC 9651](https://datatracker.ietf.org/doc/html/rfc9651#section-2) has a lot of stuff in it.  I've boiled all that down as much as I think we can get away with here.

vasilvv: I am trying to understand how this went from tokens to tokens and strings; our implementation currently only accepts tokens, for what it's worth.

martenrichter: Maybe a stupid question, but how are values separated? By commas as for WebSocket?

martenrichter: And another question, I have set "WT-Available-Protocols" but google/quiche complains:
```
E0406 12:18:14.509503   84630 quic_spdy_stream.cc:1870] Invalid character in header name WT-Available-Protocols
```
reason is that quiche only allows lower case, so should it be written in lower case?

martinthomson: Values are comma-separated, as defined in RFC 9651.

The upper/lower thing might be a bug in Google's implementation.  @DavidSchinazi @vasilvv are better positioned to answer that than I.

DavidSchinazi: A few points:
* HTTP header names[^1] are case-insensitive
* HTTP/2 and HTTP/3 require header names to be transmitted in lower-case
* The HTTP style guide recommends using mixed case for them in specifications

Because of all this, the spec should still say "WT-Available-Protocols" but if you're inside QUICHE you should pass in "wt-available-protocols". General-purpose embedders of quiche (like Chrome and Envoy) and have a conversion layer where they lowercase the header names before passing them in the quiche

[^1]: on the wind, I hear Mark whispering _fields! they're fields!_

martenrichter: Thanks! That clarifies it, and I hope to finish implementing this feature in the library next weekend.

vasilvv: We've already implemented parsing, but it only accepts tokens (not strings), which is why I am trying to understand why we're changing this.

DavidSchinazi: @martinthomson so can we change this text to say Tokens and be done with it?

martinthomson: I should have said, RFC 9651 says:

> Note that Tokens are defined largely for compatibility with the data model of existing HTTP fields and may require additional steps to use in some implementations. As a result, new fields are encouraged to use Strings.

Aside from protecting Victor's implementation (which can change, last I checked), is there a reason not to accept strings?  I say that because there are a range of valid protocol identifiers that can't be carried in tokens.  Those include anything with uppercase letters (strangely, the registry has none of those; even HTTP/1.1 is identified in lowercase).

I understand why we wouldn't be able to *remove* tokens, but it seems like there's no reason to prohibit strings, where strings are what it should have been originally.

draft-ietf-webtrans-http3.md
```diff
@@ -293,11 +293,20 @@ SHALL include a single choice from the client's list in that field. Servers MAY
...
 Both `WT-Available-Protocols` and `WT-Protocol` are Structured Fields
-{{!RFC8941}}. `WT-Available-Protocols` is a List of Tokens, and `WT-Protocol` is
-a Token. The token in the `WT-Protocol` response header field MUST be one of
-the tokens listed in `WT-Available-Protocols` of the request.  The semantics of
-individual token values is determined by the WebTransport resource in question
-and are not registered in IANA's "ALPN Protocol IDs" registry.
+{{!FIELDS=RFC9651}}. `WT-Available-Protocols` is a List, with the only valid
+value types being String or Token.  `WT-Protocol` is defined as an Item with the
+only valid value types being String or Token. Any value type other than String
+or Token MUST be ignored.  String or Token values can be used interchangeably,
```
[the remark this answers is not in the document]
martinthomson: Fair question.  It was a coin toss on my end.  An advantage of ignoring is that you get to add extensions that aren't protocol identifiers.  A disadvantage of ignoring is that you get to add extensions that aren't protocol identifiers.

The easiest alternative would be to treat the presence of anything else as an error that causes the entire field to be dropped.  Which do you prefer?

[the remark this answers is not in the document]
martinthomson: OK, I'll update it.

draft-ietf-webtrans-http3.md
```diff
@@ -293,11 +293,20 @@ SHALL include a single choice from the client's list in that field. Servers MAY
...
+but endpoints SHOULD use String; see {{Section 3.3.4 of !FIELDS}}.  Decoded
+values are used to identify protocols. No semantics are defined for parameters
+on either field; parameters MUST be ignored.
+
+The value in the `WT-Protocol` response header field MUST be one of
+the values listed in `WT-Available-Protocols` of the request.
```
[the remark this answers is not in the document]
martinthomson: `"bar"` is the same value as `bar`.  I'll mention that more explicitly.

draft-ietf-webtrans-http3.md
```diff
@@ -294,11 +294,23 @@ SHALL include a single choice from the client's list in that field. Servers MAY
...
+{{!FIELDS=RFC9651}}. `WT-Available-Protocols` is a List.  `WT-Protocol` is
+defined as an Item. In both cases, the only valid value types are String or
+Token. Any value type other than String or Token MUST be treated as an error
+that causes the entire field to be ignored.  String or Token values can be used
+interchangeably, but endpoints SHOULD use String; see {{Section 3.3.4 of
+!FIELDS}}.  Decoded values are used to identify protocols. No semantics are
+defined for parameters on either field; parameters MUST be ignored.
+
+The value in the `WT-Protocol` response header field MUST be one of the values
+listed in `WT-Available-Protocols` of the request.  Otherwise, the `WT-Protocol`
```
LPardue, line 306: This seems to be a restatement (albeit a better one) of the requirement on L293

>  If the server receives such a header, it MAY include a
`WT-Protocol` field in a successful (2xx) response. If it does, the server
SHALL include a single choice from the client's list in that field.

Should we just remove the earlier one?

draft-ietf-webtrans-http3.md
```diff
@@ -294,11 +294,23 @@ SHALL include a single choice from the client's list in that field. Servers MAY
...
+field MUST be ignored.  Note that this applies to the decoded value; a server
+can respond with a Token where the client lists a String, even though String
+values are preferred.
```
LPardue, line 309: dare I ask about percent-encoding?

draft-ietf-webtrans-http3.md
vasilvv: Note that the first MAY is not redundant (you're allowed to not send any response).

vasilvv (approved):

draft-ietf-webtrans-http3.md
LPardue: The question was motivated by the ugliness of Alt-Svc, and I think the issue is above structured fields. I know this proposal is **not** using ALPN IDs but I think that leaves the door open to people picking values that could hit interop issues.

Let me use https://httpwg.org/specs/rfc7838.html#rfc.section.3 to illustrate. The Alt-Svc header is defined with ABNF as 

```
Alt-Svc       = clear / 1#alt-value
clear         = %s"clear"; "clear", case-sensitive
alt-value     = alternative *( OWS ";" OWS parameter )
alternative   = protocol-id "=" alt-authority
protocol-id   = token ; percent-encoded ALPN protocol name
alt-authority = quoted-string ; containing [ uri-host ] ":" port
parameter     = token "=" ( token / quoted-string )
````

in normal operations, you're sending ALPN IDs of `h2` or `h3`. But in our infinite wisdom people lso defined `http/0.9, `http/1.0` and `http/1.1` https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids and so you have this horrible situation where you're sending `http%2F1.1` around. That meant RFC 7838 added all this gumpf

> ALPN protocol names are octet sequences with no additional constraints on format. Octets not allowed in tokens ([[RFC7230]](https://httpwg.org/specs/rfc7838.html#RFC7230), [Section 3.2.6](https://svn.tools.ietf.org/svn/wg/httpbis/specs/rfc7230.html#field.components)) MUST be percent-encoded as per [Section 2.1](https://www.rfc-editor.org/rfc/rfc3986.html#section-2.1) of [[RFC3986]](https://httpwg.org/specs/rfc7838.html#RFC3986). Consequently, the octet representing the percent character "%" (hex 25) MUST be percent-encoded as well.

> In order to have precisely one way to represent any ALPN protocol name, the following additional constraints apply:

    Octets in the ALPN protocol name MUST NOT be percent-encoded if they are valid token characters except "%", and
    When using percent-encoding, uppercase hex digits MUST be used.

> With these constraints, recipients can apply simple string comparison to match protocol identifiers.

My concern is that you're allowing endpoints to pick either Token or String and telling them WT-Protocol `MUST be one of the values` but are we talking exact wire match or logical match? For example, if I have a client that sends 

```
WT-Available-Protocols: "ch@t", "babble"
```

and a server that responds

```
ch%40t
```

will the subprotocol negotation succeed?

vasilvv:
Can we just ban tokens, and use strings?

LPardue (approved):

vasilvv (approved):

PR #216, opened 2025-08-05
In case we decide to enable flow control always

### LINKED ISSUE #214: Explicit Negotiation of Flow Control
It is unfortunate that it's not possible to use flow control when only using a single WebTransport session (SETTINGS_WT_MAX_SESSIONS = 1). Flow control would be useful in that case as well, since the WebTransport session shares the underlying QUIC connection with HTTP/3, and one might still starve the other.

My preference would go towards explicitly enabling flow control, either using a separate HTTP/3 setting, or by requiring the server to declare at least one of the initial flow control values (SETTINGS_WT_INITIAL_MAX_STREAMS_UNI, SETTINGS_WT_INITIAL_MAX_STREAMS_BIDI or SETTINGS_WT_INITIAL_MAX_DATA). For performance reasons, I'd expect any flow control-enabled WebTransport server to define this values anyway.

martinthomson:
This is the easiest solution.

Closes #214 in a way that is contrary to the conclusion at IETF 123.

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -598,31 +598,16 @@ streams.  A QUIC stream's data limit controls the amount of data that can be
 sent on that stream, WebTransport or otherwise (see {{Section 4
 of !RFC9000}}).
 
-## Negotiating the Use of Flow Control {#flow-control-negotiate}
+A WebTransport session MUST enable flow control.  This prevents an application
+from consuming excessive resources on a single session and starving traffic for
+other sessions (see {{security-considerations}}).
 
-A WebTransport endpoint that allows a WebTransport session to share an
-underlying transport connection with other WebTransport sessions MUST enable
-flow control.  This prevents an application from consuming excessive resources
-on a single session and starving traffic for other sessions
-(see {{security-considerations}}).
-
-An endpoint indicates that it is willing to support more than one WebTransport
-session, and thus flow control, by sending the SETTINGS_WT_MAX_SESSIONS with a
-value greater than "1".  If either endpoint sends SETTINGS_WT_MAX_SESSIONS with
-a value of "1", flow control is not enabled, and clients MUST NOT attempt to
-establish more than one simultaneous WebTransport session.  A server that
-receives more than one session on an underlying transport connection when flow
-control is not enabled MUST reset the excessive CONNECT streams with a
-`H3_REQUEST_REJECTED` status (see {{flow-control-limit-sessions}}).
-
-If both endpoints send SETTINGS_WT_MAX_SESSIONS with a value greater than "1",
-flow control is enabled, and the limits described in the entirety of
-{{flow-control}} apply.
-
-If flow control is not enabled, an endpoint MUST ignore receipt of any flow
-control capsules (see {{flow-control-capsules}}), since the peer might not have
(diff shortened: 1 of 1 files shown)


draft-ietf-webtrans-http3.md
```diff
@@ -598,31 +598,16 @@ streams.  A QUIC stream's data limit controls the amount of data that can be
...
+A WebTransport session MUST enable flow control.  This prevents an application
+from consuming excessive resources on a single session and starving traffic for
+other sessions (see {{security-considerations}}).
+
+Note that the flow control limits set by a client might not be received by a
+server before it receives a request to start a WebTransport session.  In that
+case, the newly arrived settings might establish non-zero limits for sessions
+({{flow-control-limit-sessions}}) or streams ({{flow-control-limit-streams}}).
+These can be treated as equivalent to the receipt of a WT_MAX_SESSIONS or
+WT_MAX_STREAMS frame, respectively.
```
marten-seemann, line 610: It's even simpler: Before handling a WebTransport request, the server MUST check that it's not malformed (see section 3.1). Among others, this is the case if the client didn't enable HTTP/3 datagrams. This means that the server has to wait for the HTTP/3 SETTINGS before processing a WebTransport request.

marten-seemann (approved):
I like making flow-control non-optional. Always enabling flow control simplifies the protocol quite a bit.

draft-ietf-webtrans-http3.md
martinthomson: I don't think that's right.  There's a paragraph that applies to [draft versions only](https://ietf-wg-webtrans.github.io/draft-ietf-webtrans-http3/draft-ietf-webtrans-http3.html#section-3.1-7):

> For draft versions of WebTransport only, the server MUST NOT process any incoming WebTransport requests until the client settings have been received

In a full version, the receipt of a request to start a session is sufficient for a server to know that the client supports webtransport.  What it lacks is the ability to know what settings are in place, which might be essential for operation.  This text merely notes that these will eventually arrive and that it is OK.

(Text like this is necessary, in my view, no matter what we decide on the flow control issue.)

marten-seemann: That's inconsistent with the requirement to check for HTTP/3 datagram support:
> Any WebTransport requests sent by the client without enabling QUIC and HTTP datagrams MUST be treated as malformed by the server, as described in [Section 4.1.2](https://rfc-editor.org/rfc/rfc9114#section-4.1.2) of [[HTTP3](https://ietf-wg-webtrans.github.io/draft-ietf-webtrans-http3/draft-ietf-webtrans-http3.html#HTTP3)].

Should we open a new issue to clarify? I'd be ok with converting this to a SHOULD, or having the server wait for SETTINGS (in the common case, SETTINGS will arrive in the same flight as the WebTransport request).

martenrichter:
Another question, if flow-control is optional, how would backpressure of the streams be implemented, which is needed on the W3C side?

kixelated:
> Another question, if flow-control is optional, how would backpressure of the streams be implemented, which is needed on the W3C side?

QUIC has flow control. The idea is that when a QUIC connection is dedicated to WebTransport, which I imagine will be the case the vast majority of the time, all flow control is delegated to the QUIC layer.

Issue #238, opened 2025-11-03
Early IANA review summary


LPardue:
IANA performed an early review of the draft, to summarize some of the points

1. Section 9.2 - Please use the field names used in the HTTP/3 settings registry. "Code" instead of "Value", "Initial Value" instead of "Default". _CHAIR NOTE: this seems incorrect based on https://www.iana.org/assignments/http3-parameters/http3-parameters.xhtml#http3-parameters-settings and RFC 9114_
2. Section 9.3, 9.4 - The HTTP/3 frame tyypes and stream types use "Value" instead of "Code". We also need "Status, Change Controller" and "Contact" fields
3. Section 9.5 - The HTTP/3 error codes need "Status, Change Controller" and "Contact" fields

PR #240, opened 2025-11-03
Address some early IANA feedback

LPardue:
This partly address some feedback in #238. It doesn't touch on the "Status" fields, which are inconsisently absent or present with permanent or provisional. I'll make a separate PR for that.


ekinnear (approved):
Thank you!

PR #242, opened 2025-11-03
Separate out settings prose and registration

LPardue:
SETTINGS_WT_MAX_SESSIONS already had prose in section 3.1 that was not consistently
duplicated in 9.2, so I just deleted it.

Prose about SETTINGS for flow control is moved to section 5.5 which was already
called "Flow Control SETTINGS".

The first version of this PR nukes some of the anchor magic because it seemed a bit pointless to me. But if people feel strongly then it can be added back.

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -744,16 +744,50 @@ capsule as a session error.
 
 ## Flow Control SETTINGS
 
-*[SETTINGS_WT_INITIAL_MAX_STREAMS_UNI]: #
-*[SETTINGS_WT_INITIAL_MAX_STREAMS_BIDI]: #
-*[SETTINGS_WT_INITIAL_MAX_DATA]: #
-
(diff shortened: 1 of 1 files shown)


draft-ietf-webtrans-http3.md
```diff
@@ -744,17 +744,51 @@ capsule as a session error.
...
+WT_MAX_STREAMS capsule (see {{WT_MAX_STREAMS}}).  The default value for the
+SETTINGS_WT_INITIAL_MAX_STREAMS_UNI setting is "0", indicating that the
+endpoint needs to send WT_MAX_STREAMS capsules on each individual WebTransport
+session before its peer is allowed to create any unidirectional streams within
+that session.
+
+Note that this limit applies to all WebTransport sessions that use the HTTP/3
+connection on which this SETTING is sent.
+
+### SETTINGS_WT_INITIAL_MAX_STREAMS_BIDI
```
martinthomson: Don't we need `_local` and `_remote` versions of this?

PR #268, opened 2026-06-01
Editorial fixes from Alan's WGLC review

ekinnear:
* Reword "Stream independence can mitigate head-of-line blocking" and replace "unopinionated" with "without prescribing their usage".
* Replace "is not used for anything else related to an established session" with "is only used for minimal framing".
* Use "any necessary header fields" rather than "perhaps some" or "possibly some" in the minimal-handshake bullets.
* Drop "in general" from the server-identification paragraph.
* Use "a CONNECT stream" rather than "the CONNECT stream" for agreement.
* Replace descriptive "may" with "can" or "might", and convert SHALL to MUST for the buffered-stream/datagram drop, GOAWAY follow-up, and TLS exporter cases.
* Use "WebTransport sessions in 0-RTT packets" rather than "WebTransport in 0-RTT packets".
* Use "possible next protocols" in the WT-Available-Protocols description, mirroring TLS ALPN.
* Fix "(8..)" to "(..)" for the Exporter Label length encoding so it matches the Context field below it.
* Drop "so" from "and so are prohibited".
* Drop "considered" from "is considered terminated".

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -83,11 +83,11 @@ accept incoming WebTransport sessions.
 ## QUIC, WebTransport, and HTTP/3
 
 QUIC version 1 {{!RFC9000}} is a secure transport protocol with flow control and
-congestion control.  QUIC supports application data exchange via streams;
-reliable and ordered byte streams that can be multiplexed.  Stream independence
-can mitigate head-of-line blocking.  While QUIC provides streams as a transport
-service, it is unopinionated about their usage.  The applicability of streams is
-described by section 4 of {{?RFC9308}}.
+congestion control.  QUIC supports application data exchange via streams:
+reliable and ordered byte streams that can be multiplexed.  These streams are
+independent, so head-of-line blocking does not propagate between them.  QUIC
+provides streams as a transport service without prescribing their usage.  The
+applicability of streams is described by section 4 of {{?RFC9308}}.
 
 HTTP is an application-layer protocol, defined by "HTTP Semantics" {{HTTP}}.
 HTTP/3 is the application mapping for QUIC, defined in {{!RFC9114}}.  It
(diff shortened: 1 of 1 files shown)


draft-ietf-webtrans-http3.md
```diff
@@ -83,11 +83,11 @@ accept incoming WebTransport sessions.
...
-congestion control.  QUIC supports application data exchange via streams;
-reliable and ordered byte streams that can be multiplexed.  Stream independence
-can mitigate head-of-line blocking.  While QUIC provides streams as a transport
-service, it is unopinionated about their usage.  The applicability of streams is
-described by section 4 of {{?RFC9308}}.
+congestion control.  QUIC supports application data exchange via streams:
+reliable and ordered byte streams that can be multiplexed.  These streams are
+independent, so head-of-line blocking does not propagate between them.  QUIC
+provides streams as a transport service without prescribing their usage.  The
+applicability of streams is described by section 4 of {{?RFC9308}}.
```
martinthomson:
```suggestion
applicability of streams is described by {{Section 4 of ?RFC9308}}.
```

draft-ietf-webtrans-http3.md
```diff
@@ -103,9 +103,8 @@ is important for security reasons, especially to ensure that the resource is
 willing to use WebTransport.
 
 Although WebTransport requires HTTP for its handshake, when HTTP/3 is in use,
-HTTP is not used for anything else related to an established session.  Instead,
-QUIC streams begin with a sequence of header bytes that links them to the
-established session.  The remainder of the stream is the body, which carries the
+HTTP is only used for minimal framing.  QUIC streams begin with a sequence of
```
martinthomson:
```suggestion
HTTP is used for session setup and only minimal framing beyond that.
QUIC streams begin with a sequence of
```

draft-ietf-webtrans-http3.md
```diff
@@ -655,7 +654,7 @@ defined in {{Section 7.5 of !RFC8446}} with the label set to
 WebTransport Exporter Context {
   WebTransport Session ID (64),
   WebTransport Application-Supplied Exporter Label Length (8),
-  WebTransport Application-Supplied Exporter Label (8..),
+  WebTransport Application-Supplied Exporter Label (..),
```
martinthomson: I think that this is wrong.  The exporter label has to be at least one byte in length.  Otherwise all exporter contexts will be equal.

```suggestion
  WebTransport Application-Supplied Exporter Label (8..),
```

PR #269, opened 2026-06-01
Deeper updates from Alan's WGLC review

### LINKED ISSUE #262: WGLC Review
The Abstract could use a rewrite that leads with “This document …”

The terminology paragraph flows a little awkwardly as well.

> Stream independence can mitigate head-of-line blocking.

This sentence feels stilted

> it is unopinionated about their usage

Is this adding value?  Also not a word/misspelled.

> HTTP/3 offers two features in addition to HTTP Semantics: QPACK header compression [RFC9208] and Server Push Section 4.6 of [RFC9114].

Do we need to mention either in this document?  Certainly server push can be dropped?

> is not used for anything else related to an established session

Perhaps: “is only used for minimal framing”

> and perhaps some extra header fields.

And any necessary header fields

> and possibly some extra header fields

> A WebTransport endpoint, whether a client or a server, can likely perform several of its HTTP-level requirements using bytestring comparisons.

? I’m not sure what we’re suggesting here.  I recall there was a lot of push back about “h3 is hard” and this section seems to be saying “its not that hard”.  Do we still need this?

> 2.1.2. Capsule-Based WebTransport over HTTP/3

Is this section in the right place?  Maybe should come after overview?

> Endpoints SHOULD always use this protocol

Do we need a reason why one might not?

>  in general are identified by a pair

Drop “in general”?

> The ID of the CONNECT stream

“a CONNECT stream” has better agreement

> f it does not, it SHOULD reply with status code 404

But overview security considerations has:

> For example, the client must not be able to distinguish between a network address that is unreachable and one that is reachable but is not a WebTransport server.

>  WebTransport server MAY accept the session

This is a funny MAY - perhaps “can” or “server accepts”

> server may reply with a 3xx response, indicating a redirection (

Lowercase may

> Clients cannot initiate WebTransport in 0-RTT packets,

WebTransport sessions

> SETTINGS parameters may be retained

Lowercase may

> If the server accepts 0-RTT, the server MUST NOT reduce the limit of maximum open WebTransport sessions,

I think we dropped max sessions

> such change would be deemed incompatible

Because these values are incompatible, the server MUST send …

> enumerates the possible protocols 

Possible next protocols

> If it does, the server MUST include a single choice from the client's list in that field

Otherwise, request error or connection error?  Aha, it’s a couple paragraphs later - 

>  If the client receives a WT-Protocol value that was not included in its WT-Available-Protocols list, the client MUST close the WebTransport session with a WT_ALPN_ERROR error code.

> SHALL be 0x54.

We’re still planning to renumber into the 1 byte varint space, right?

> A WebTransport endpoint may send a 

Lowercase may

> similarly, intermediaries MUST reset such streams with a corresponding error code when receiving a reset from their peer. 

Is this talking about STOP_SENDING -> RST_STREAM behavior?  If so, it’s too vague.

> If a RESET_STREAM or STOP_SENDING frame is received with an error code outside the range reserved for WT_APPLICATION_ERROR, the stream is still considered reset, but the error code is not mapped to a WebTransport application error code. The WebTransport implementation SHOULD deliver this to the application as a stream reset with no application error code.

This is super annoying.  Essentially your interface now needs to have optional<uint32_t> error code on the off chance that the peer sent a broken code.  Can we reserve the MAX error code in WT space as INTERNAL_ERROR or some such, and say that unknown errors map to INTERNAL_ERROR?  I will file this as a separate issue.

> a WebTransport server could be put into a situation where it receives a stream or a datagram without a corresponding session.

Server can receive

>  client may receive

Lowercase may

> a stream SHALL be

Are we consistent with MUST/SHALL?

> it must open a new

Lowercase must

> and MAY open new WebTransport streams.

Mention datagrams too?

> when the backend shuts down

Backend is used without definition. Might need more explicit text

> QUIC connection may be shared

Lowercase may

> SHALL be a TLS exporter

Shall

>   WebTransport Application-Supplied Exporter Label (8..),

What does (8..) mean?  Is it a typo?

> endpoints can limit the number of sessions that a peer can create on a single HTTP/3 connection

Still true?  Via rate limit header or maybe we just drop it

>  are not used and so are prohibited. 

Drop ‘so’ ?

[throughout - need to update to reference RFC’s of things that have shipped e.g. HTTP-DATAGRAM?

> type of 0x190B4D3F

Renumbering here I assume

> HTTP/3 is considered terminated 

is terminated

> the endpoint MUST reset the send side and abort reading on the receive side of all unidirectional and bidirectional streams associated with the session (see Section 2.4 of [RFC9000]) using the WT_SESSION_GONE error code;

What error code does the implementation deliver to the application when a stream is reset with WT_SESSION_GONE?

ekinnear:
Each change is in its own commit to make it easy to back out or adjust any of them.

* Restructure the Abstract so the document subject opens the paragraph
* Smooth the Terminology paragraph flow
* Remove the QPACK and Server Push aside
* Remove the bytestring-comparison note
* Drop "always" from "SHOULD use this protocol"
* Drop the max-sessions clause from 0-RTT settings and reword the 
* Forward-reference the ALPN error path, so the consequence of 
* Be explicit about intermediary RESET_STREAM and STOP_SENDING forwarding. 
* Note that datagrams can also be sent after WT_DRAIN_SESSION/GOAWAY
* Use "origin server" rather than "backend"
* Make explicit that connection-level session count is governed by 
* Fold session-gone stream resets into the session-termination event 

Closes #262

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -358,12 +358,14 @@ intent is to simplify porting existing protocols that use QUIC and rely on this
 functionality.
 
 The client MAY include a `WT-Available-Protocols` header field in the CONNECT
+The client MAY include a `WT-Available-Protocols` header field in the CONNECT
 request.  The `WT-Available-Protocols` field enumerates the possible next
 protocols in preference order, with the most preferred protocol listed first.
 If the server receives such a header, it MAY include a `WT-Protocol` field in
 a successful (2xx) response.  If it does, the server MUST include a single
-choice from the client's list in that field.  Servers MAY reject the request
-if the client did not include a suitable protocol.
+choice from the client's list in that field; client behavior when the server
+response does not satisfy this requirement is described below.  Servers MAY
+reject the request if the client did not include a suitable protocol.
 
 Both `WT-Available-Protocols` and `WT-Protocol` are Structured Fields
 {{!FIELDS=RFC9651}}.  `WT-Available-Protocols` is a List.  `WT-Protocol` is
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -142,9 +142,6 @@ header field, and any necessary header fields.
 * Generating/parsing the response status code and any necessary header
   fields.
 
-A WebTransport endpoint, whether a client or a server, can likely perform
-several of its HTTP-level requirements using bytestring comparisons.
-
 While HTTP/3 encodes HTTP messages using QPACK, this complexity can be
 minimized.  When receiving, a WebTransport endpoint can disable dynamic
 decompression entirely but must always support static decompression and Huffman
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -95,9 +95,7 @@ HTTP is an application-layer protocol, defined by "HTTP Semantics" {{HTTP}}.
 HTTP/3 is the application mapping for QUIC, defined in {{!RFC9114}}.  It
 describes how QUIC streams are used to carry control data or HTTP request and
 response message sequences in the form of frames and describes details of stream
-and connection lifecycle management.  HTTP/3 offers two features in addition to
-HTTP Semantics: QPACK header compression {{?RFC9208}} and Server Push
-{{Section 4.6 of RFC9114}}.
+and connection lifecycle management.
 
 WebTransport session establishment involves interacting at the HTTP layer with a
 resource.  For Web user agents and other WebTransport clients, this interaction
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -68,15 +68,17 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
 when, and only when, they appear in all capitals, as shown here.
 
-This document follows terminology defined in Section 1.2 of [OVERVIEW].  Note
-that this document distinguishes between a WebTransport server and an HTTP/3
-server.  An HTTP/3 server is the server that terminates HTTP/3 connections; a
-WebTransport server is an application that accepts WebTransport sessions, which
-can be accessed via an HTTP/3 server.  An application client is user or
-developer-provided code, often untrusted, that utilizes the interface offered by
-the WebTransport client to communicate with an application server. The
-application server uses the interface offered by the WebTransport server to
-accept incoming WebTransport sessions.
+This document follows terminology defined in Section 1.2 of [OVERVIEW].
+WebTransport servers and HTTP/3 servers are distinguished here as two separate
+roles: an HTTP/3 server terminates HTTP/3 connections, while a WebTransport
+server is an application that accepts WebTransport sessions, accessed via an
+HTTP/3 server, possibly through zero or more intermediaries.
+
+An application client is user-provided or developer-provided code, often
+untrusted, that uses the interface offered by the WebTransport client to
+communicate with an application server.  The application server uses the
+interface offered by the WebTransport server to accept incoming WebTransport
+sessions.
 
 # Overview
 
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -1093,7 +1093,11 @@ Upon learning that the session has been terminated, the endpoint MUST reset the
 send side and abort reading on the receive side of all unidirectional and
 bidirectional streams associated with the session
 (see {{Section 2.4 of !RFC9000}}) using the WT_SESSION_GONE error code;
-it MUST NOT send any new datagrams or open any new streams.
+it MUST NOT send any new datagrams or open any new streams.  WT_SESSION_GONE is
+a protocol-level error code rather than an application error code; the
+WebTransport implementation SHOULD report these stream resets to the
+application as part of the session-termination event rather than as separate
+stream-reset events carrying the WT_SESSION_GONE code.
 
 To terminate a session with a detailed error message, an application MAY provide
 such a message for the WebTransport endpoint to send in an HTTP capsule
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -332,10 +332,10 @@ Clients cannot initiate WebTransport sessions in 0-RTT packets, as the CONNECT
 method is not considered safe (see {{Section 10.9 of HTTP3}}).  However,
 WebTransport-related SETTINGS parameters can be retained from the previous
 session as described in {{Section 7.2.4.2 of HTTP3}}.  If the server accepts
-0-RTT, the server MUST NOT reduce the limit of maximum open WebTransport
-sessions, or other initial flow control values, from the values negotiated
-during the previous session; such change would be deemed incompatible, and MUST
-result in a H3_SETTINGS_ERROR connection error.
+0-RTT, the server MUST NOT reduce these initial flow control values from those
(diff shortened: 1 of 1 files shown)


draft-ietf-webtrans-http3.md
```diff
@@ -68,15 +68,17 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
-This document follows terminology defined in Section 1.2 of [OVERVIEW].  Note
-that this document distinguishes between a WebTransport server and an HTTP/3
-server.  An HTTP/3 server is the server that terminates HTTP/3 connections; a
-WebTransport server is an application that accepts WebTransport sessions, which
-can be accessed via an HTTP/3 server.  An application client is user or
-developer-provided code, often untrusted, that utilizes the interface offered by
-the WebTransport client to communicate with an application server. The
-application server uses the interface offered by the WebTransport server to
-accept incoming WebTransport sessions.
+This document follows terminology defined in Section 1.2 of [OVERVIEW].
```
martinthomson:
```suggestion
This document follows terminology defined in {{Section 1.2 of OVERVIEW}}.
```

draft-ietf-webtrans-http3.md
```diff
@@ -68,15 +68,17 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+WebTransport servers and HTTP/3 servers are distinguished here as two separate
+roles: an HTTP/3 server terminates HTTP/3 connections, while a WebTransport
+server is an application that accepts WebTransport sessions, accessed via an
+HTTP/3 server, possibly through zero or more intermediaries.
```
martinthomson:
```suggestion
HTTP/3 server, possibly through an intermediary.
```

draft-ietf-webtrans-http3.md
```diff
@@ -68,15 +68,17 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+
+An application client is user-provided or developer-provided code, often
+untrusted, that uses the interface offered by the WebTransport client to
+communicate with an application server.  The application server uses the
+interface offered by the WebTransport server to accept incoming WebTransport
+sessions.
```
martinthomson: Do you want to mention that the WebTransport client could be a web browser, where the application is the website running in that browser?  Seems like a useful piece of information to share.

draft-ietf-webtrans-http3.md
```diff
@@ -68,15 +68,17 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
```
martinthomson, line 91: When the text in the document says just "client" or "server", which of the different client or server roles does that refer to?  It might help to explain that it generally refers to the WebTransport client.

But the fact that a WebTransport server might be on the other side of an intermediary REALLY complicates things there, because the rules in the protocol need to be implemented in the intermediary.  So you have to be super careful with that server definition.

draft-ietf-webtrans-http3.md
```diff
@@ -153,8 +150,8 @@ static compression, or Huffman encoding.
 ### Capsule-Based WebTransport over HTTP/3
 
 WebTransport over HTTP/3 as defined in this document provides the best
-performance by using native QUIC streams and datagrams. Endpoints SHOULD always
-use this protocol when using WebTransport over an HTTP/3 connection.
+performance by using native QUIC streams and datagrams. Endpoints SHOULD use
```
martinthomson:
```suggestion
performance by using native QUIC streams and datagrams. Endpoints SHOULD prefer
```

?

martinthomson (commented):
I saved comments, but failed to share them.  So you get those.  They might be old, but they didn't look like it.

draft-ietf-webtrans-http3.md
ekinnear: Good point. Added a note about assuming it's WebTransport client and WebTransport server. 

I took a pass at explaining the intermediary part, very open to suggestions: 
> An intermediary between a WebTransport client and an upstream WebTransport
server presents itself as a WebTransport server to the WebTransport client,
and as a WebTransport client to the upstream WebTransport server.
Requirements this document places on WebTransport servers therefore apply to
intermediaries when acting as a server, and requirements on WebTransport
clients apply when acting as a client.

draft-ietf-webtrans-http3.md
ekinnear: Added!

martinthomson (approved):
A little clumsy perhaps with all the WebTransport this and WebTransport that, but workable.

draft-ietf-webtrans-http3.md
```diff
@@ -682,10 +694,11 @@ zero-length if omitted.
 
 Flow control governs the amount of resources that can be consumed or data that
 can be sent.  When using WebTransport over HTTP/3, endpoints can limit the
-number of sessions that a peer can create on a single HTTP/3 connection and the
-number of streams that a peer can create within a session.  Endpoints can also
-limit the amount of data that can be consumed by each session and by each stream
-within a session.
+number of sessions that a peer can create on a single HTTP/3 connection via
+rate limiting (see {{flow-control-limit-sessions}}), and the number of streams
```
afrind, line 700: Why 'rate limiting'  when referencing flow control?

draft-ietf-webtrans-http3.md
```diff
@@ -1087,7 +1100,11 @@ Upon learning that the session has been terminated, the endpoint MUST reset the
 send side and abort reading on the receive side of all unidirectional and
 bidirectional streams associated with the session
 (see {{Section 2.4 of !RFC9000}}) using the WT_SESSION_GONE error code;
-it MUST NOT send any new datagrams or open any new streams.
+it MUST NOT send any new datagrams or open any new streams.  WT_SESSION_GONE is
+a protocol-level error code rather than an application error code; the
+WebTransport implementation SHOULD report these stream resets to the
+application as part of the session-termination event rather than as separate
+stream-reset events carrying the WT_SESSION_GONE code.
```
afrind: I think I get what we're trying to say -- basically indicating that the peer application didn't send these resets so the application can distinguish, but are we being too heavy handed in describing how the implementation/API should work?

draft-ietf-webtrans-http3.md
```diff
@@ -575,9 +585,10 @@ a WebTransport data stream.  This ensures reliable delivery of the ID field
 associating the data stream with a WebTransport session.
 
 WebTransport endpoints MUST forward the error code for a stream associated with
-a known session to the application that owns that session; similarly,
-intermediaries MUST reset such streams with a corresponding error code when
-receiving a reset from their peer.  If a RESET_STREAM or STOP_SENDING frame is
+a known session to the application that owns that session.  Upon receiving a
+RESET_STREAM or STOP_SENDING frame on a WebTransport stream, an intermediary
+MUST send the same frame type, with the corresponding error code, to the next
```
afrind: Perhaps the paragraph would hold together better with a topic sentence like "The error code send by the application in abrupt termination messages is part of the WebTransport API and MUST NOT be altered or suppressed by implementations or intermediaries.

afrind (approved):

draft-ietf-webtrans-http3.md
ekinnear: Since we're specifically using the rate limiting mechanism from httpapi, seemed helpful to be consistent here

draft-ietf-webtrans-http3.md
ekinnear: Updated!

draft-ietf-webtrans-http3.md
ekinnear: Updated to be a bit more descriptive and less proscriptive

PR #280, opened 2026-07-02
WebTransport is enabled with a "1"

martinthomson:
Only.  Other values result in `H3_SETTINGS_ERROR`.

This is consistent with `SETTINGS_ENABLE_CONNECT_PROTOCOL`, where the only valid values are 0 or 1.

I've also added a mention of the setting have no effect when sent by a client.

Closes #275.

### DIFF
--- a/draft-ietf-webtrans-http3.md
+++ b/draft-ietf-webtrans-http3.md
@@ -221,7 +221,7 @@ This document defines a SETTINGS_WT_ENABLED setting that WebTransport servers
 use to indicate their support for WebTransport.  The default value for the
 SETTINGS_WT_ENABLED setting is "0", meaning that the server does not support
 WebTransport.  A value of "1" indicates support for the variant of WebTransport
-that is described in this document (that is, "webtransport-h3").  Clients MAY
+that is described in this document (that is, "webtransport-h3").  Clients MUST
 treat values greater than "1" as a connection error of type H3_SETTINGS_ERROR.
 Clients MUST NOT attempt to establish WebTransport sessions with the
 "webtransport-h3" token until they have received the setting indicating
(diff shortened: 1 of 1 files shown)


draft-ietf-webtrans-http3.md
```diff
@@ -220,8 +220,12 @@ establishing sessions (see {{upgrade-token}}).
 This document defines a SETTINGS_WT_ENABLED setting that WebTransport servers
 use to indicate their support for WebTransport.  The default value for the
 SETTINGS_WT_ENABLED setting is "0", meaning that the server does not support
-WebTransport.  Clients MUST NOT attempt to establish WebTransport sessions until
-they have received the setting indicating WebTransport support from the server.
+WebTransport.  A value of "1" indicates support for the variant of WebTransport
+that is described in this document (that is, "webtransport-h3").  Clients MAY
```
marten-seemann:
```suggestion
that is described in this document (that is, "webtransport-h3").  Clients MUST
```

marten-seemann (commented):
I'm wondering if we should keep higher values as an extension point, but it's probably easier to just a different SETTINGS value for that.

draft-ietf-webtrans-http3.md
martinthomson: The reason I say "MAY" is that clients might not bother checking settings that they don't plan on using.  For instance, a client might not check the setting until a WebTransport session is requested.  Our code wouldn't be able to generate an error, for instance.  (At least, not without some reworking.)

marten-seemann: If you don't understand the setting, it's totally fine to ignore it. But if you understand it, you MUST validate it. At least that's how we deal with it on the QUIC side, for example if a transport parameter be set to 0 or 1.

martinthomson: Sure.  Also, I just found where we check the value and we already reject values >1.  So MUST is fine with me.

ekinnear (approved):

